PR #19, opened 2026-05-02
housekeeping: shrink avatar_images.cc from 2.73MB to ~6KB (closes #9)

### LINKED ISSUE #9: Shrink `avatar_images.cc` placeholder (currently 2.86MB of zeros)
## Background

\`firmware/main/boards/stackchan/avatar_images.cc\` is a placeholder: 14 RGB565 image symbols (160×120) where every byte is `0x00`. Because each symbol is 38400 bytes serialized as `0x00,0x00,...` text, the file is **2.86MB on disk** even though the compressed git blob is much smaller.

The placeholder exists so the firmware links and boots without shipping any real avatar art. Users are expected to replace it with their own.

## Goal

Make the placeholder file itself small so the repo doesn't show a 2.86MB scary-looking file.

## Approach options

1. **1×1 RGB565 per symbol** — 14 × 2 bytes = 28 bytes total. Requires verifying that `stackchan.cc`'s display logic doesn't crash when handed a 1×1 `lv_image_dsc_t` (it should just render a single black pixel).
2. **CONFIG_LV_USE_RLE on a single black byte** — preserves the (160, 120) header but uses RLE-compressed data, getting ~14 × ~10 bytes total.
3. **Generator script + .gitignore** — keep the placeholder out of git, regenerate on first build via `convert_avatars.py --generate-placeholder`. Cleanest, most setup work.

Option 1 is the simplest. Verify with: `idf.py build` succeeds and the device boots without crash; LCD just shows the default xiaozhi UI behind whatever LVGL renders.

## Acceptance

- `avatar_images.cc` is under 10KB on disk
- `idf.py build` succeeds for the stackchan board
- Device boots without crashing

Tanayqq:
Replaced 14 × 160×120 zero-filled RGB565 arrays with 14 × 1×1 black-pixel lv_image_dsc_t entries. File shrinks from 2.73MB to ~6KB.

Implements Option 1 from the issue.


kisaragi-mochi:
Thanks @Tanayqq, merged! 🙏

Verified end-to-end on M5Stack CoreS3 + StackChan hardware before merging:

- `python ./scripts/release.py stackchan` builds clean, no new warnings on the changed file
- `xiaozhi.bin` flashes and boots normally (no LVGL complaints, PSRAM init OK, all MCP tools registered including `self.display.set_avatar` / `set_mouth` / `set_blink`)
- StateMachine reaches `listening`, gateway sees `connected=true` with all 14 tools
- `set_avatar("happy")` / `set_avatar("sad")` / `set_blink(true)` all return `applied=1` end-to-end — LVGL accepts the 1×1 source and the 2x scale path runs without error (the placeholder pixel is simply too small to be visible against the chat UI emoji underneath, which is expected for a placeholder)

Resolves #9. The file size is now ~7 KB (down from 2.73 MB) and the boot path is unchanged.

One thing surfaced during verification that's unrelated to your change: the `CONFIG_DEFAULT_WEBSOCKET_URL` fallback added in #27 doesn't fire on devices that previously ran upstream xiaozhi-esp32 firmware (NVS still holds `wss://api.tenclass.net/...`). I've opened #28 to add a force-override switch so future contributors testing on ex-xiaozhi hardware don't get stuck in the same loop. Mentioning it here in case you ever flash to a board with stale NVS yourself.

Thanks again for the clean implementation and for picking this one up!

Tanayqq:
“Thanks for the detailed verification and for merging! Great to know it works end-to-end on the hardware. Also appreciate you flagging the NVS fallback issue—I’ll watch out for that in future testing.”

PR #68, opened 2026-05-09
feat(LED): add MCP tools for the 12x WS2812C base RGB strip

Saariba:
## Summary

Note: this PR was created with help of Claude code.

Add MCP control of the 12× WS2812C RGB LEDs on the StackChan base.

The strip is wired to the **PY32L020 IO expander** on its pin 13 (not an
ESP32 GPIO), sharing the AW88298 / Si12T I2C bus that the firmware
already talks to for `VM_EN` (servo power) and head-touch. The firmware
extends the existing in-tree `Py32IoExpander` helper with the same LED
API the M5 BSP exposes — `SetDriveMode`, `SetLedCount`, `SetLedColor`
(RGB888 → RGB565 packing per `((r&0xF8)<<8)|((g&0xFC)<<3)|(b>>3)`),
`SetLedData` (single-burst write of all 12 RGB565 pairs), and
`RefreshLeds` (read-modify-write so the count nibble survives the
bit-6 latch trigger). `InitializeIOExpander` configures pin 13 as
push-pull output with pull-up after the existing VM_EN block, sets
count = 12, waits the M5-prescribed 200 ms before the first refresh,
and clears the strip. An `rgb_ok_` flag tracks init success so the
tools degrade with `available=false` instead of cascading errors when
the PY32 is unreachable.

Four new ESP32 MCP tools, each implicitly latching so updates are
WYSIWYG:

- `self.led.set_color {index, r, g, b}`
- `self.led.set_all {r, g, b}`
- `self.led.set_many {colors}` (JSON-encoded `[[r,g,b], ...]`)
- `self.led.clear`

Four matching gateway-side surfaces with clean names:

- `set_led(index, r, g, b)`
- `set_all_leds(r, g, b)`
- `set_leds(colors)` — accepts a real JSON array on the gateway side
  and re-packs as a JSON string for the device, since the on-device
  MCP property layer only supports scalar types
- `clear_leds()`

The batch tool exists so animations / multi-color patterns are one
I2C burst + one refresh instead of 12 round-trips.

`docs/architecture.md` tool-mapping table updated; `CHANGELOG.md`
gains an `Added` entry under `[Unreleased]`.

### Code-level

- [x] gateway: `uv run pytest` (84 passed)
- [x] gateway: `uv run ruff check .` (clean)
- [x] firmware: run all checks and 

### Hardware

- [x ] Device boots without crash
- [ x] Existing MCP tools still work where affected: `move_head`,
      `take_photo`, `set_volume`, `get_head_angles`
- [ x] Existing touch/servo behavior still works where affected: tap,
      stroke, wobble
- [ x] New behavior verified on real hardware: run `set_all_leds(255,0,0)`,
      `set_led(6, 0, 255, 0)`, `set_leds([[255,0,0],[0,255,0],[0,0,255]])`,
      `clear_leds()` and confirm the 12-LED base ring follows each
      command with no flicker on unrelated indices.
- [ x] Confirm `get_status` still reports the device as connected and
      the four new tools appear in the device's discovery list.

## Breaking changes

None. The four new tools are purely additive on both the gateway and
the device. Existing tool names and semantics are unchanged.

## Related issues

None — opening as a standalone enhancement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,32 @@ change is called out under a `Firmware` subsection of the release entry.
 
 ## [Unreleased]
 
+### Added
+
+- New MCP tools to drive the 12× WS2812C RGB LEDs on the StackChan
+  base: `set_led(index, r, g, b)`, `set_all_leds(r, g, b)`,
+  `set_leds(colors)` (batch, single I2C burst for animations), and
+  `clear_leds()`. The strip is wired to the PY32L020 IO expander on
+  expander pin 13 — not an ESP32 GPIO — so the firmware extends the
+  existing `Py32IoExpander` helper with `SetDriveMode`, `SetLedCount`,
+  `SetLedColor` (RGB888 → RGB565 packing), `SetLedData` (burst write),
+  and `RefreshLeds` (RMW preserves the count nibble alongside the
+  bit-6 latch trigger), matching the M5 BSP's
+  `PY32IOExpander_Class.cpp`. `InitializeIOExpander` configures pin 13
+  as push-pull output with pull-up, sets the LED count to 12, waits
+  the M5-prescribed 200 ms before the first refresh, then clears the
+  strip. The on-device tools (`self.led.set_color`, `self.led.set_all`,
+  `self.led.set_many`, `self.led.clear`) all latch implicitly so each
+  call is WYSIWYG; the gateway re-packs the Python `colors` list as a
+  JSON string for `set_many` since the device-side MCP property layer
+  has only scalar types. `set_many` validates every entry (including
+  a `cJSON_IsNumber` guard so non-number elements like `"255"` or
+  `null` no longer silently coerce to 0) before any I2C write, so
+  malformed input cannot leave the strip in a partially-mutated
+  state. All four tools no-op cleanly with an `available=false`
+  reply when the PY32 init failed, so a flaky expander does not
+  cascade into errors.
+
 ## [0.4.0] - 2026-05-09
 
 ### Firmware
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,26 @@ change is called out under a `Firmware` subsection of the release entry.
 
 ### Added
 
+- New MCP tools to drive the 12× WS2812C RGB LEDs on the StackChan
+  base: `set_led(index, r, g, b)`, `set_all_leds(r, g, b)`,
+  `set_leds(colors)` (batch, one I2C burst for animations), and
+  `clear_leds()`. The strip is wired to the PY32L020 IO expander on
+  expander pin 13 — not an ESP32 GPIO — so the firmware extends the
+  existing `Py32IoExpander` helper with `SetDriveMode`, `SetLedCount`,
+  `SetLedColor` (RGB888 → RGB565 packing), `SetLedData` (burst write),
+  and `RefreshLeds` (RMW preserves the count nibble alongside the
+  bit-6 latch trigger), matching the M5 BSP's
+  `PY32IOExpander_Class.cpp`. `InitializeIOExpander` configures pin 13
+  as push-pull output with pull-up, sets the LED count to 12, waits
+  the M5-prescribed 200 ms before the first refresh, then clears the
+  strip. The on-device tools (`self.led.set_color`, `self.led.set_all`,
+  `self.led.set_many`, `self.led.clear`) all latch implicitly so each
+  call is WYSIWYG; the gateway re-packs the Python `colors` list as a
+  JSON string for `set_many` since the device-side MCP property layer
+  has only scalar types. All four tools no-op cleanly with an
+  `available=false` reply when the PY32 init failed, so a flaky
+  expander does not cascade into errors.
+
 - The on-device WiFi configuration UI now also exposes a **Fallback
   Gateway URL** field and a **Gateway Token** field on the **Advanced**
   tab, alongside the existing WebSocket Gateway URL field. The values
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -52,6 +52,10 @@ The gateway translates between two tool naming conventions:
 | `set_blink` | `self.avatar.set_blink` |
 | `set_mouth` | `self.avatar.set_mouth` |
 | `check_vm_en` | `self.robot.check_vm_en` |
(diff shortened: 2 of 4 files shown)


kisaragi-mochi (commented):
Thanks for this — really nice PR. The design follows the existing
`Py32IoExpander` pattern cleanly, the `rgb_ok_` flag mirrors `servo_ok_`
for graceful degradation, and the doc updates (architecture.md mapping +
CHANGELOG entry) are exactly right. Reading through the firmware side,
I have two findings I'd love your help on, plus a small ops note:

### Findings

**[P2] `set_leds` (batch) does not use the burst writer**
`firmware/main/boards/stackchan/stackchan.cc:2108-2111`

The tool docstring and PR description both advertise "one I2C burst",
but the inner loop calls `SetLedColor` once per LED, so a 12-LED update
is 12 round-trips plus the refresh — roughly equivalent to
`set_color × N`. Could you pack the RGB565 pairs into a local buffer
and call `SetLedData` once before `RefreshLeds()`? That would let
`set_leds` actually deliver the latency property the API was designed
for, which matters once animation code starts using it.

The "missing entries keep their previous color" semantics make the
partial-write case slightly tricky — you'd need to either readback
first, or restrict the burst length to the supplied count and document
that trailing LEDs aren't touched. Happy to defer to whichever you
prefer.

**[P2] Non-numeric color channels are silently accepted**
Same location.

Because `colors` arrives as a JSON-encoded string, the typed property
validation is bypassed. `cJSON` leaves `valueint` as 0 for elements
like `"255"` (string) or `null`, so a malformed payload such as
`[["255", 0, 0]]` currently writes black to LED 0 and returns
`ok=true` instead of the documented error. Adding a
`cJSON_IsNumber(jr) && cJSON_IsNumber(jg) && cJSON_IsNumber(jb)`
check before clamping should surface the existing
`"Each entry must be a [r,g,b] triple of integers"` error path
correctly.

### Ops notes

- The PR is showing conflicts because `CHANGELOG.md`'s `[Unreleased]`
  got promoted to `[0.4.0]` after release v0.4.0, and
  `docs/architecture.md` picked up a `set_mouth_sequence` row from
  PR #66. The code changes here don't overlap with anything that
  landed since `41865ea`, so `git rebase main` should resolve cleanly
  with just those two doc files. Let me know if you'd prefer the
  maintainer side handles the rebase instead.
- CI hasn't run yet — that's just GitHub's first-time-contributor
  gate; will approve and trigger workflows shortly.

Thanks again — looking forward to the next iteration!

kisaragi-mochi:
**Real-device validation update**

Just took this branch for a spin on a physical StackChan (worktree on `refs/pull/68/head`, my Tailscale Funnel URL in `sdkconfig.defaults.local`). All four new tools work end-to-end via the gateway:

- `set_all_leds(255,0,0)` → all 12 LEDs solid red ✅
- `set_leds([[255,0,0],[0,255,0],[0,0,255]])` → index 0/1/2 = R/G/B, 3..11 retained previous color (matches the documented "missing entries keep their previous color" semantics) ✅
- `clear_leds()` → all off ✅
- 12-LED rotating-rainbow and knight-rider sweeps via repeated `set_leds` — visually correct, no dropped frames at the agent-driven rate.

A couple of refinements to my earlier review based on what I saw:

**Finding 1 (per-LED I2C round-trips)**: in practice the loop-based path is ≈3–4 ms for 12 LEDs at 100 kHz, so it is **not visually distinguishable** from a single-burst `SetLedData()` path. The visual-correctness framing in my original review was stronger than warranted — apologies for that. The suggestion still stands, but on weaker grounds:

1. **Doc / code consistency** — the inline comment on `SetLedData()` says *"Single I2C transaction — much faster than calling SetLedColor in a loop"*, and the sibling `set_all` / `clear` paths already burst. `set_many` should follow.
2. **Latency budget margin** — fine in isolation, but the burst path leaves headroom that the loop does not for any future high-frequency caller.

**Finding 2 (cJSON validation)**: not exercised by the real-device test, because the gateway's pydantic schema rejects non-integer `colors` upstream. The firmware-side `cJSON_IsNumber()` check is still worth adding as defense-in-depth for any future caller that bypasses the gateway, but it's nice-to-have rather than blocking.

**Out-of-scope follow-up**: the agent-driven frame rate is structurally bounded to hundreds of ms per frame by the LLM ↔ MCP ↔ WS ↔ I2C chain. Smooth (≥24 fps) animation needs firmware-side primitives. Tracked separately as #69 — future enhancement, not a blocker for this PR.

Thanks again for the clean implementation; the new tools are a solid foundation. Looking forward to the rebased follow-up.

Saariba:
Thanks for your tests! Good catches. 

I adressed both of your findings: 
1. `set_many` is now a real burst and does not contradict with the comment.

2. cJSON_IsNumber` guard added** before the `valueint` reads. The malformed-payload error path (`"Each entry must be a [r,g,b] triple of integers"`)

I also rebased it as well. 



I agree with you that it is a good idea to track separately the issue seperatly https://github.com/kisaragi-mochi/stackchan-mcp/issues/69. 


I tested it again my local StackChan and it worked with no issues
Thank you!

kisaragi-mochi (approved):
**Fix verified — approving** ✅

Both original findings are addressed cleanly, and the atomic-validation improvement is a nice extra step beyond what I flagged.

- **Finding 1 (set_many burst)** — packs the validated prefix into a local 24-byte buffer, then `SetLedData` + `RefreshLeds` once. Matches the sibling `set_all` / `clear` paths and the `SetLedData` inline comment. ✅
- **Atomicity bonus** — validation runs over the whole array before any I2C write, so a malformed entry at i=5 can no longer leave LEDs 0..4 mutated. The `written` semantic shift (now `ok ? n : 0` instead of a partial count) is consistent with that — there is no longer a "partial success" state to report.
- **Finding 2 (cJSON_IsNumber)** — string / null / bool can no longer silently coerce to 0 and write black while reporting `ok=true`. ✅
- **n=0 success path** — defense-in-depth for direct device callers that bypass the gateway's `minItems=1` schema. ✅
- **CHANGELOG re-add** — thanks for catching the v0.4.0 main-merge drop and for refreshing the wording to mention the new burst + atomic semantics.

CI: Gateway test ✅ (12 s); Firmware build ✅ (5m59s) — this is good to merge from my side.

Thanks for the careful follow-up — looking forward to having this merged.

PR #72, opened 2026-05-09
docs(readme): add LED tools to README tool tables

Saariba:
## Summary

Follow-up to #68. The four LED tools landed on `main` and are
documented in `docs/architecture.md` and `CHANGELOG.md`, but the
user-facing tool tables in the three READMEs were not updated. This
PR adds matching rows so the LEDs show up alongside the other tools
in the entry-point docs.

- `README.md` (English root) — 4 rows added at the end of the table
- `README.ja.md` (Japanese root) — 4 matching rows, kept in sync per
  CONTRIBUTING.md's "top-level user guide is maintained in both
  English and Japanese" rule
- `gateway/README.md` — 4 rows plus a short note that all four LEDs
  share the PY32 I2C bus and degrade with `available=false` when
  the expander init fails (this file is the per-tool detail
  reference, so it gets the longer descriptions; the `set_leds`
  entry calls out the single-burst + atomic-validation properties)

No code changes.

### Code-level

- [ ] gateway: `uv run pytest` — not run (no behavior change)
- [ ] gateway: `uv run ruff check .` — not run (no Python files
      changed)
- [ ] firmware: `python ./scripts/release.py stackchan` — not
      applicable (docs only)
- [x] Not applicable / not run: docs-only PR, no code changes

### Hardware

- [x] Not applicable / hardware not available: docs-only change

## Breaking changes

None.

## Related issues

Refs #68.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -57,6 +57,10 @@
 | `set_mouth(state)` | 口開閉（one-shot、次の呼び出しまで保持） | ✅ |
 | `set_mouth_sequence(steps)` | TTS リップシンク用に `{shape, duration_ms}` のリストをデバイス側でキュー再生（ステップごとの WebSocket RTT ゆらぎなし） | ✅ |
 | `check_vm_en` | サーボ電源 (VM EN HIGH) 状態確認 | ✅ |
+| `set_led(index, r, g, b)` | ベース部の RGB LED 12 個のうち 1 個を指定 (index `0..11`、各チャネル `0..255`) | ✅ |
+| `set_all_leds(r, g, b)` | ベース部の RGB LED 12 個すべてを同じ色に設定 | ✅ |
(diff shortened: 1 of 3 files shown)


kisaragi-mochi (approved):
**Doc follow-up — approving** ✅

Beautifully clean follow-up. Three doc files in sync, with the right level of detail for each:

- **`README.md` / `README.ja.md`** — concise tool table rows that match the existing pattern, kept in sync per the CONTRIBUTING dual-language rule. ✅
- **`gateway/README.md`** — fuller per-tool descriptions, including the `set_leds` "single I2C burst + atomic validation" properties, and a short note on the shared PY32 I2C bus + `available=false` degradation. Exactly the right place for those details. ✅

CI: Gateway test ✅ / Firmware build ✅ — good to merge.

Thanks for the careful follow-up — these doc updates were on my list and you saved me the work 🙏

PR #87, opened 2026-05-11
feat(firmware,gateway): #85 envelope-driven TTS lip-sync with state-cycle fallback

### LINKED ISSUE #85: feat(tts): drive mouth lip-sync from per-frame audio amplitude envelope (Phase 4 polish, follow-up to #76)
## Background

PR #84 (Issue #76) shipped the **(A) state-driven** approach for TTS
lip-sync: while the gateway is playing TTS audio (\`tts.start\`..
\`tts.stop\`), the firmware cycles the mouth through
\`closed → half → open → half\` on a fixed 150 ms cadence. This already
lifts the device from \"talks with mouth shut\" to \"talks with mouth
moving\" with no audio analysis required.

This issue tracks the **(B) audio-envelope-driven** follow-up
proposed in the original Issue #76: replace the fixed cadence with a
per-frame audio amplitude mapping so the mouth opens wider on louder
syllables and stays mostly closed during quiet stretches, producing a
more natural appearance.

## Proposal

Map a per-frame audio amplitude envelope (RMS over each ~60 ms Opus
frame) to a mouth shape, then drive the existing
\`SetMouthShape()\` path with that mapping instead of the fixed cycle:

- \`rms < threshold_low\` → \`closed\`
- \`threshold_low ≤ rms < threshold_high\` → \`half\`
- \`threshold_high ≤ rms\` → \`open\`

Thresholds should be chosen to match the dynamic range of the engines
shipped today (VOICEVOX TTS in PR #75, Irodori in the upcoming PR3
slice) and ideally calibrated automatically per-utterance so a
quieter voice does not stay perpetually closed.

## Implementation options

There are two natural places to compute the envelope; either can
replace the (A) state machine:

- **Firmware-side post-decode**: hook into the existing audio decode
  pipeline (\`audio_service_.PushPacketToDecodeQueue\`) and compute
  RMS on each PCM frame after Opus decode. Lowest latency
  (envelope-to-mouth happens entirely on the device), but hard-wires
  the threshold logic into firmware and requires re-flashing to tune.

- **Gateway-side precompute**: compute RMS per Opus frame in the
  gateway's TTS pipeline (where the raw PCM is already available
  before encoding), serialise the envelope alongside the audio
  stream, and send a sidecar JSON message
  (e.g. \`{\"type\":\"tts.envelope\",\"frame_id\":N,\"rms\":...}\`)
  before each audio frame. Firmware just maps incoming RMS values to
  mouth shapes via \`SetMouthShape()\`. More flexible (thresholds and
  mapping can be tuned per-engine in the gateway without re-flashing,
  and engines that already emit phoneme timing could feed it through
  the same channel later) but adds protocol surface and per-frame
  network overhead.

The gateway-side option is probably the better default because it
keeps engine-specific calibration in the gateway, but the firmware-
side option is closer to today's code path. Recommend prototyping
both small enough to compare on the same utterance.

## Hooking into the (A) implementation

PR #84 added the lip-sync state machine in
\`firmware/main/boards/stackchan/stackchan.cc\` as a single
\`esp_timer\`-driven cycle gated on \`tts_lipsync_active_\`. The
machinery for **starting / stopping the animation, pausing /
restoring blink, and yielding to user-issued \`set_mouth_sequence\`**
is the part this issue should keep — only the per-step shape
selection (\`TtsLipSyncStepAdvance()\`'s \`switch\` statement) needs
to change to consume an envelope value rather than walk a fixed
cycle.

## Out of scope

- **Phoneme-aligned lip-sync** (Japanese mora / vowel category
  timing). VOICEVOX's \`/audio_query\` returns \`mora\` /
  \`accent_phrases\` JSON that could feed this, but Irodori (PR3)
  ships only bare audio, so a uniform cross-engine approach starting
  from the amplitude envelope keeps the firmware path single-track.
  Phoneme-driven lip-sync is a separate further follow-up.

## Refs

- #76 — Phase 4 lip-sync (A) tracking issue
- #84 — (A) state-driven implementation (Closes #76)
- #75 — \`say()\` pipeline that surfaced the missing binding
- #70 — Phase 4 audio parent
- #65 — Blink end repaints active mouth overlay (related avatar
  rendering concern; unchanged by this work)

kisaragi-mochi:
## Status

**On hold.** Initial implementation pushes an envelope sidecar per Opus frame. On-device testing on the Tailscale-WSS deployment path showed the per-frame `ws.send()` pair (envelope JSON + Opus binary) consumes more than the 60 ms pacing budget, producing audible audio underruns. Sub-sampling the envelope cadence (tested at N=3, N=5, N=10) only stops the underruns at cadences that exceed the firmware's 200 ms `ENVELOPE_FRESH_WINDOW_US`, at which point the envelope-driven mouth path becomes unreachable while the audio risk remains.

The state-driven fallback merged in #84 already drives a usable mouth animation, so this PR stays open as a placeholder pending a wire-format redesign that avoids the two-send-per-frame contention.

Likely paths forward:
- Embed the envelope into the same WebSocket binary frame as the Opus payload (single send per audio frame, no JSON parse cost on device).
- Move the envelope onto a separate side-channel that does not contend with the audio frame send lock.

Closes #85 (when one of the redesign paths is in scope).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,25 +15,55 @@ change is called out under a `Firmware` subsection of the release entry.
 
 ## [Unreleased]
 
+### Gateway
+
+- **TTS audio amplitude envelope sidecar (Issue #85)**: the
+  orchestrator now emits one
+  `{"type":"tts","state":"envelope","frame_id":N,"rms":...}` JSON
+  message per Opus audio frame, carrying the per-frame RMS amplitude
+  (normalised to `[0.0, 1.0]`) computed from the same PCM that fed
+  the encoder. Firmware on the StackChan board (see Firmware section)
+  consumes this to drive amplitude-based mouth-shape selection during
+  lip-sync; older firmware ignores the message and falls back to the
+  fixed lip-sync cycle introduced in the previous release. Each
+  envelope is awaited inline before the matching audio frame so the
+  WebSocket write order is guaranteed at the connection layer (a new
+  per-connection `_send_lock` on `ESP32Connection` serialises every
+  underlying `websockets.send()`). To keep that ordering from
+  inflating the 60 ms audio pacing budget, the envelope path uses a
+  bounded `lock_acquire_timeout=30 ms` — if another sender already
+  holds the send lock past that window, the envelope is dropped
+  silently and the audio frame proceeds on schedule, with the
+  firmware fallback cycle covering the missed update. The bound is
+  applied at the *lock acquire* layer rather than around `ws.send()`
+  itself because the `websockets` library documents cancelling an
+  in-flight send as unsafe. Refs #76, #84.
+
 ### Firmware
 
-- **TTS lip-sync (state-driven)**: drive avatar mouth animation while
-  the gateway is speaking. The firmware now reacts to the
-  `tts.start` / `tts.stop` JSON notifications introduced in #75 (Issue
-  #70 PR2) and cycles the mouth shape through `closed → half → open →
-  half` on a fixed 150 ms cadence for the lifetime of each utterance,
-  snapping back to `closed` at stop. Autonomous blink is paused while
-  active (same Phase 2 trade-off as the existing `set_mouth_sequence`
-  task: a blink ending would otherwise restore the full-face image
-  and overwrite the mouth overlay) and restored at stop based on
-  `blink_desired_` so a `set_blink` issued mid-playback is honoured.
-  Coexists with user-issued `set_mouth` / `set_mouth_sequence` calls
-  by yielding the current frame when `mouth_seq_active_` is true; the
-  user-issued sequence wins until it completes, then lip-sync resumes
-  on the next tick. Wired through a new no-op `Board::OnTtsStart` /
-  `OnTtsStop` hook so non-stackchan boards are unaffected. The (B)
-  audio-envelope-driven follow-up proposed in the issue will be
-  tracked separately. Closes #76. Refs #70, #75.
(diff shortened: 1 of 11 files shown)


kisaragi-mochi:
Withdrawing this PR.

End-to-end testing on the actual deployment path (Tailscale-based WSS to the device) showed that per-frame envelope sidecars compounded with audio frames consume more than the 60 ms pacing budget, producing audible audio underruns on the device. Sub-sampling envelopes (tested at N=3, N=5, N=10) borderline-resolved the audio glitch only at cadences that exceeded the firmware's 200 ms `ENVELOPE_FRESH_WINDOW_US` — i.e. the firmware fell back to its fixed-cycle path regardless, so the envelope-driven mouth path was no longer reachable while the underrun risk remained.

The state-driven fallback already merged in #84 drives a satisfactory mouth animation, so the additional WS-level protocol overhead introduced here is not justified at this point.

Issue #85 stays open for a future redesign that either embeds the envelope into the WS binary frame payload (single send per audio frame, no JSON parse cost on the device) or uses a separate transport that does not contend with the audio frame send lock. The foundational pieces (per-connection send lock, ``lock_acquire_timeout`` API, ``compute_pcm_frame_rms`` helper, firmware envelope hook on ``Board::OnTtsEnvelope``) are preserved in this branch's history for that follow-up.

kisaragi-mochi:
Closing again — the design-level tracking lives in #85, so this PR does not need to stay open as a placeholder. A fresh PR will be opened when a redesign path (binary-frame embedded envelope, separate side-channel, etc.) is in scope.

Issue #70, opened 2026-05-09
feat(tts): gateway-side say() — Phase 4 slice with no firmware changes


kisaragi-mochi:
## Goal

Slice the TTS half of #8 so the gateway drives the device speaker via a new `say(text)` MCP tool — TTS runs on the gateway side, output is Opus-encoded 16 kHz mono, pushed to the device over the existing WebSocket. **No firmware changes.**

## Why slice TTS first

- Half of #8 in scope, **zero firmware churn** (the device-side xiaozhi-esp32 protocol already accepts Opus audio frames; handshake is in place via `gateway/stackchan_mcp/esp32_client.py`).
- Immediately useful: an LLM agent can speak from the device with one new tool, before the STT loop lands.
- Engine plug-in design lets STT (which still blocks on #8) evolve independently.

### MCP tool surface

```python
say(
  text: str,
  voice: str = "voicevox",            # engine selector
  speaker_id: int | None = None,       # VOICEVOX
  reference_audio: str | None = None,  # Irodori voice-cloning reference
)
```

### Engine abstraction

```
gateway/stackchan_mcp/tts/
├─ base.py       # Engine ABC: synthesize(text, **opts) -> bytes (PCM 16k mono)
├─ voicevox.py   # default — VOICEVOX HTTP API
└─ irodori.py    # optional — Irodori-TTS-500M (MIT, zero-shot voice cloning)
```

### Extras

- `pip install stackchan-mcp[tts]` — base (httpx, opuslib)
- `pip install stackchan-mcp[tts-voicevox]` — VOICEVOX (no extra deps; requires a running VOICEVOX server)
- `pip install stackchan-mcp[tts-irodori]` — Irodori-TTS (torch, transformers — heavier)

Users who only want vision / servo / LED control pay zero TTS dependency cost.

### Engine choice

| Engine | Default | License | Why |
|---|---|---|---|
| **VOICEVOX** | ✅ | LGPL-3.0 (HTTP server, process-isolated — no copyleft contamination) | Most popular Japanese TTS; easy Docker deploy; well-documented |
| **Irodori-TTS-500M** | optional | MIT | Zero-shot voice cloning from a reference audio sample — the agent can have its own voice. Heavier dependency footprint. |

Other engines (AivisSpeech / Style-Bert-VITS2 / OpenAI TTS / Google TTS) can land in follow-up PRs once the engine abstraction is in place.

## Acceptance

[...]

Issue #64, opened 2026-05-09
WebSocket connection reset after ~9s when reaching the gateway via Tailscale Funnel


kisaragi-mochi:
## Summary

When the firmware connects to the stackchan-mcp gateway through a Tailscale Funnel `wss://` URL (`wss://<node>.<tailnet>.ts.net/`, per `docs/remote-access.md`), the WebSocket session is reset by the peer roughly 9 seconds after the post-handshake `connecting -> listening` transition, even when the device is idle and the gateway is healthy. Direct LAN deployments (`ws://<lan-ip>:8765/`) are unaffected.

## Reproduction

1. Build firmware with `CONFIG_FORCE_DEFAULT_WEBSOCKET_URL=y` and `CONFIG_DEFAULT_WEBSOCKET_URL="wss://<node>.<tailnet>.ts.net/"`
2. Run `tailscale funnel --bg --https=443 http://localhost:8765` so the Funnel forwards `:443` to the local gateway
3. Boot the ESP32 and trigger a WS connection (LCD tap or wake word)
4. Observe `Connected to websocket server` → `connecting -> listening` → ~9 seconds later, `Websocket disconnected (no reconnect: candidate failed or intentional close)`

## Serial log fingerprint

```
I (...) WS: Connected to websocket server candidate 1/1: wss://<node>.<tailnet>.ts.net/
I (...) StateMachine: State: connecting -> listening
... ~9s ...
E (...) Dynamic Impl: mbedtls_ssl_fetch_input error=76
E (...) esp-tls-mbedtls: read error :-0x004C
E (...) EspSsl: SSL receive failed: -76
I (...) WS: Websocket disconnected (no reconnect: candidate failed or intentional close)
```

`-0x004C = MBEDTLS_ERR_NET_CONN_RESET` — the remote side initiated the close.

## Likely causes (to investigate)

- Tailscale Funnel's HTTP/HTTPS-oriented idle reaper closing long-lived WebSocket upgrade connections; Funnel is HTTPS-first and may not be tuned for the WS upgrade lifecycle.
- Gateway-side WebSocket `ping_interval` / `ping_timeout` (Python `websockets` defaults) not aligned with the Funnel's idle timeout, so server-initiated keep-alive pings either don't fire or aren't routed back through the Funnel within the expected window.
- A middlebox between the two collapsing the underlying TCP socket once application traffic stops.

## Impact

[...]

PR #93, opened 2026-05-12
feat(stt): #91 gateway-side listen() MCP tool with minimal firmware listen trigger

### LINKED ISSUE #91: feat(stt): gateway-side listen() — Phase 4 slice with no firmware changes
## Goal

Slice the STT half of #8 so the gateway can route incoming microphone audio from the device through a Speech-to-Text engine and expose the transcription as a new `listen()` MCP tool. **Minimal firmware change** to add a server-driven `listen` wire-type handler — see PR #93 for the design rationale.

Companion slice to #70 (which carried the TTS half forward).

## Why slice STT now

- Half of #8 in scope, **minimal firmware change (~30 lines in `Application::OnIncomingJson`)** — the firmware does not stream microphone audio unsolicited (the original premise here was wrong), so a small additive `listen` wire-type handler was needed to drive the device into listening mode. The handler is symmetric to the existing device→gateway `Protocol::SendStartListening` notification. See PR #93 for the design discussion.
- Immediately useful for builders who want a desktop companion that talks back: an LLM agent can capture a short user utterance via one new tool, without waiting on streaming / VAD work.
- Pairs naturally with #70 (`say()`): a round-trip conversation becomes possible once both are merged.
- Engine plug-in design lets STT and TTS evolve independently.

### MCP tool surface

Start with single-shot capture; streaming can come later.

```python
listen(
  duration_ms: int = 5000,           # capture window
  engine: str = \"faster-whisper\",    # engine selector
  language: str | None = \"ja\",       # ISO 639-1; None = auto-detect
  model: str | None = None,          # engine-specific (e.g. \"base\" / \"small\" / \"medium\")
) -> { text: str, duration_ms: int, language: str }
```

Optional VAD-trimmed capture (defer to follow-up if it complicates the slice):

```python
listen_until_silence(
  max_duration_ms: int = 10000,
  silence_ms: int = 800,
  ...
) -> { text, duration_ms, language }
```

### Engine abstraction

Mirrors the TTS layout from #70:

```
gateway/stackchan_mcp/stt/
├─ base.py             # Engine ABC: transcribe(pcm16k_mono: bytes, **opts) -> { text, language }
├─ faster_whisper.py   # default — faster-whisper, local
└─ openai_whisper.py   # optional — OpenAI Whisper API
```

### Audio pipeline reuse

- Decode incoming Opus 16 kHz mono → PCM, using the same opuslib path that #70 will need for outgoing audio.
- Buffer `duration_ms` of frames into a single PCM blob and hand it to the engine.
- Build on the existing `audio_stream.py` stub and `esp32_client.py` handshake.

### Extras

- `pip install stackchan-mcp[stt]` — base (opuslib, numpy)
- `pip install stackchan-mcp[stt-faster-whisper]` — faster-whisper (CPU/GPU local)
- `pip install stackchan-mcp[stt-openai]` — OpenAI Whisper API (cloud, user supplies `OPENAI_API_KEY`)

Users who only want vision / servo / LED / TTS pay zero STT dependency cost.

### Engine choice

| Engine | Default | License | Why |
|---|---|---|---|
| **faster-whisper** | ✅ | MIT | Local, no cloud round-trip, Japanese accuracy is solid, CPU-runnable on a recent Mac mini. Default to match \"runs offline out of the box\" stance. |
| **OpenAI Whisper API** | optional | (cloud, user's own API key) | Lowest setup friction for users without local compute; works on a Raspberry Pi gateway host. |

Other engines (Vosk for ultra-low latency, whisper.cpp for tiny footprint, Azure / Google STT) can land in follow-up PRs once the engine abstraction is in place.

## Latency target

Match #8's acceptance bar: under 2 seconds for a short Japanese utterance with the local default engine on a typical developer Mac.

For the OpenAI Whisper API path, latency is dominated by upload + cloud RTT and is documented rather than gated.

## Acceptance

- [ ] New MCP tool `listen(duration_ms)` works end-to-end with the default `faster-whisper` engine: user speaks into the device microphone → transcription string returned to the caller, with no firmware changes.
- [ ] `listen(duration_ms, engine=\"openai-whisper\")` works for users who install the `stt-openai` extra and have an API key configured.
- [ ] Under 2 seconds for a short Japanese utterance with the local default engine (measured on a recent Mac mini).
- [ ] Engine abstraction (`stt/base.py`) is symmetric to #70's `tts/base.py` so the two layers feel consistent.
- [ ] Documentation: how to install each extra, where the API key goes, sample latency numbers, language autodetect notes.
- [ ] `CHANGELOG.md` `[Unreleased]` Gateway entry.

## Out of scope

- **Streaming / interim transcription.** Single-shot capture only in this slice; streaming + VAD is a follow-up.
- **Wake-word detection.** Separate concern, not in this issue.
- **Continuous conversation loop.** Composing `listen()` + LLM + `say()` into a loop is the LLM agent's responsibility, not the gateway's.
- Firmware changes other than the additive `listen` wire-type handler above; the existing `Application::OnIncomingAudio` / decode-queue pipeline is reused unchanged.

## Related

- Refs #8 (Phase 4 parent — TTS half landed via #70; this issue carries the STT half forward)
- Companion to #70 (TTS slice, same architecture: gateway-side engine, no firmware churn)
- Builds on existing `gateway/stackchan_mcp/audio_stream.py` stub and `esp32_client.py` handshake

kisaragi-mochi:
## Summary

Wire up the STT half of Phase 4 (#8) so the gateway can drive the device into listening mode on demand, buffer the inbound Opus microphone frames, and transcribe them through a registered STT engine. Adds a new `listen(duration_ms?, engine?, language?, model?)` MCP tool.

Closes #91.

## Why

#70 (PR #74 / #75) shipped the TTS half of Phase 4 last week — the gateway can speak through the device speaker. This PR is the symmetric inbound half: the gateway can ask the device to open its microphone, capture a short utterance, and return a transcription. Once both land, an LLM agent can hold a round-trip conversation through StackChan via just two MCP calls.

## Wire format

Mirrors the existing device→gateway `Protocol::SendStartListening` notification in the reverse direction:

```
{"session_id": "...", "type": "listen", "state": "start", "mode": "manual"}
{"session_id": "...", "type": "listen", "state": "stop"}
```

The upstream 78/xiaozhi-esp32 protocol does not currently emit `type:"listen"` server→device (confirmed by reading `Application::OnIncomingJson` upstream — it only accepts `tts` / `stt` / `llm` / `mcp` / `system` / `alert` / `custom`), and xinnan-tech/xiaozhi-esp32-server only handles inbound listen. The reverse-direction extension is additive and does not collide with anything upstream.

## Firmware change (minimal)

A new `type: "listen"` case in `Application::OnIncomingJson` dispatches `state: "start"` / `"stop"` to the existing `Application::StartListening` / `StopListening` API (~30 lines, scoped to `firmware/main/application.cc`).

The `mode` field is parsed but currently ignored — `HandleStartListeningEvent` unconditionally calls `SetListeningMode(kListeningModeManualStop)`, which is also the right behaviour for gateway-driven capture: the gateway controls the stop boundary explicitly via `state: "stop"`. Threading `auto` / `realtime` modes through is a follow-up.

This supersedes the Issue body's original "No firmware changes" framing — see the design note below.

## Gateway changes

- **`stt/` package** symmetric to `tts/`: `STTEngine` ABC + `EngineRegistry` + `orchestrator.py` (start_recording → listen.start → wait → listen.stop → stop_recording → decode → engine.transcribe) + `audio_utils.py` (Opus decode) + `faster_whisper.py` (default engine, MIT) + `openai_whisper.py` (cloud engine).
- **`audio_stream.py`**: new recording slot (`start_recording` / `stop_recording` / `is_recording`); the binary handler dispatches matching-session frames into the slot when active and discards them otherwise.
- **`esp32_client.py`**: new `send_listen_state(state, mode)` on both `ESP32Connection` and `ESP32Manager`. The `listen_lock` is an alias for `tts_lock` (single per-device audio-path lock) — see design note below.
- **`stdio_server.py`**: new `listen` MCP tool with the same clean-error discipline as `say()` (`ValueError` / `NotImplementedError` / `RuntimeError` translate to MCP error JSON).
- **`pyproject.toml`**: new optional extras `[stt]` (opuslib base), `[stt-faster-whisper]`, `[stt-openai]`.

### "No firmware changes" doesn't hold

Issue #91 proposed STT as a no-firmware-change slice on the premise that the device "already streams Opus 16 kHz mono microphone frames over the existing WebSocket." Reading the current firmware (`Application::OnIncomingAudio`, `audio_service_.EnableVoiceProcessing`, `HandleStartListeningEvent`) showed that the device only opens the microphone after a button press / wake-word — there is no gateway-driven listen trigger today. Three options were considered: (1) pure no-firmware (redefine `listen()` to "transcribe the next user-initiated utterance"), (2) minimal firmware change to add a `type:"listen"` handler, (3) re-scope the Issue.

This PR takes (2). The wire format is symmetric to the existing `SendStartListening` notification, so the firmware change is ~30 lines and stays inside `OnIncomingJson`.

### Audio-path lock

The first codex pass surfaced that an independent `listen_lock` would allow `listen()` to overlap with `say()` on the same device. `HandleStartListeningEvent` aborts in-flight TTS when listen.start arrives mid-speaking (`state == kDeviceStateSpeaking → AbortSpeaking → SetListeningMode`), so concurrent operations would have stepped on each other. The lock is now unified — `_listen_lock = _tts_lock`. If a full-duplex contract ever lands later, the lock can split again.

### Cancellation-safe stop

The first codex pass also surfaced that a cancelled `listen()` (or any exception unwinding through the capture sleep) could skip `listen.stop` and leave the firmware in `kDeviceStateListening` with the microphone open until a button press / wake-word reset it. The orchestrator now sends `listen.stop` through `asyncio.shield` inside `finally`, so the shielded send completes synchronously even when the outer await is being cancelled.

### Session-id matching on inbound frames

The second codex pass surfaced that `audio_stream.handle_audio_frame` previously appended any frame to the recording buffer regardless of session. When ESP32 reconnects, the old socket's `async for message in ws` loop can drain a binary frame or two before the close fully lands, and those stale frames would have corrupted a fresh listen() capture. The handler now compares `session_id` against the recording's owner and drops the mismatch.

### Engine-extra probe at import time

The third codex pass surfaced that both engine modules used to import their heavy ML dependencies lazily inside `transcribe()`. That meant engines registered successfully even without the matching extras, and a `listen()` call would put the device into recording mode + wait the full capture window before failing at decode time. Each engine module now probes its optional dependency at top-of-module import; the existing `_try_register` helper in `stt/__init__.py` swallows the resulting ImportError, so a missing extra surfaces as `NotImplementedError: Available engines: [...]` before any device-side state change.

## Validation

- **Unit tests**: 200 passed (3 deselected for libopus host availability — the same skip pattern already used for the TTS audio_utils tests). New coverage includes engine registry, argument validation, protocol-v1 gate, listen_lock serialisation, cancellation safety, session-id matching, empty-buffer handling, recording-slot lifecycle, and the `send_listen_state` wire format.
- **Lint**: `uv run ruff check .` passes.
- **Codex review**: 3 rounds (1 adversarial + 2 standard). All findings (1 high + 3 P2) addressed.
- **No real-device test in this PR**. End-to-end verification requires flashing the matching firmware; will be done in a follow-up session.

## Out of scope

- Streaming / interim transcription (single-shot only).
- Wake-word detection.
- Continuous conversation loop (composing `listen()` + LLM + `say()` is the agent's responsibility).
- Mode != manual (auto / realtime). Phase 1 ignores the `mode` field firmware-side.
- Vosk / whisper.cpp / cloud STT providers other than OpenAI (the framework is engine-agnostic; additional engines can land in follow-up PRs).

## Related

- Closes #91.
- Refs #8 (Phase 4 parent — TTS half landed via #74/#75; this PR carries the STT half).
- Companion to #70.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,45 @@ change is called out under a `Firmware` subsection of the release entry.
 
 ## [Unreleased]
 
+### Added
+
+- Phase 4 STT — gateway-side `listen(duration_ms?, engine?, language?,
+  model?)` MCP tool. The gateway puts the device firmware into
+  listening mode over the existing WebSocket, buffers the Opus frames
+  the device streams up during the capture window, then decodes and
+  transcribes them through a registered STT engine. The default
+  engine is **faster-whisper** (local, MIT-licensed, runs on CPU);
+  the **OpenAI Whisper API** is available as an alternative for
+  setups without local compute. Install with
+  `pip install stackchan-mcp[stt-faster-whisper]` (or `[stt-openai]`)
+  — the `[stt]` base extra pulls in `opuslib` for inbound frame
+  decoding. The framework is engine-agnostic (`stt.STTEngine` ABC +
+  `stt.EngineRegistry`, symmetric to the existing `tts` package), so
+  additional engines can be added in follow-up PRs without touching
+  the orchestration pipeline. Configure faster-whisper with
+  `STACKCHAN_FASTER_WHISPER_MODEL` (default `base`), `_DEVICE`
+  (default `cpu`), and `_COMPUTE_TYPE` (default `int8`); the OpenAI
+  engine reads `OPENAI_API_KEY` plus the optional
+  `STACKCHAN_OPENAI_WHISPER_MODEL` (default `whisper-1`). Requires a
+  paired firmware update — see the Firmware section below. Refs #91.
+
 ### Firmware
 
+- **Server-driven listening trigger** (paired with the new
+  `listen()` MCP tool above, Issue #91). The firmware's
+  `Application::OnIncomingJson` handler now accepts inbound
+  `{"type":"listen","state":"start"|"stop"}` messages from the
+  gateway and dispatches them to `Application::StartListening` /
+  `StopListening`. The wire format mirrors the existing device→gateway
+  `Protocol::SendStartListening` notification in the reverse
+  direction; the upstream 78/xiaozhi-esp32 protocol has no inbound
+  listen type today, so this extension is additive and does not
+  collide with anything upstream. The `mode` field is parsed on
+  `state="start"` but currently ignored — `HandleStartListeningEvent`
+  unconditionally enters `kListeningModeManualStop`, which is also
+  the right behaviour for gateway-driven capture (the gateway
+  controls the stop boundary explicitly). Threading `auto` /
+  `realtime` mode through is a follow-up. Refs #91.
 - **TTS lip-sync (state-driven)**: drive avatar mouth animation while
   the gateway is speaking. The firmware now reacts to the
   `tts.start` / `tts.stop` JSON notifications introduced in #75 (Issue
--- a/README.ja.md
+++ b/README.ja.md
@@ -62,6 +62,7 @@
 | `set_leds(colors)` | `[[r,g,b], ...]` 配列で先頭 N 個を一括設定（I2C 1 回のバースト送信、アニメーション等向け）。指定外の LED は前の色を保持 | ✅ |
 | `clear_leds` | ベース部の RGB LED 12 個すべて消灯 | ✅ |
 | `say(text, voice?, speaker_id?, reference_audio?)` | gateway 側 TTS でデバイススピーカーから喋らせる。デフォルトエンジンは **VOICEVOX**（別 HTTP サービスとして起動 — [TTS セットアップ](#4-オプション-tts-セットアップ-voicevox) 参照）。`[tts]` extras が必要 | ✅ |
+| `listen(duration_ms?, engine?, language?, model?)` | デバイスマイクから短い発話をキャプチャし、gateway 側 STT で文字起こし。デフォルトエンジンは **faster-whisper**（ローカル動作・MIT — [STT セットアップ](#5-オプション-stt-セットアップ-faster-whisper) 参照）。`[stt-faster-whisper]`（または `[stt-openai]`）extras と、`listen` ワイヤタイプを受け付けるファームウェアが必要 | ✅ |
 
 詳細スキーマは `gateway/README.md` 参照。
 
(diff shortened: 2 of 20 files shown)


kisaragi-mochi:
End-to-end real-device verification (follow-up to the Validation section)

## Setup

- Hardware: M5Stack CoreS3 (ESP32-S3) running the firmware built from this branch
- Flash: `xiaozhi.bin` written to `0x20000` only (NVS / WiFi / bootloader preserved)
- Transport: WebSocket protocol v1, Tailscale Funnel (`wss://...`)
- Engine: `faster-whisper` `base` model (default), `int8` compute type, CPU
- Cache status: HuggingFace model was already present (`models--Systran--faster-whisper-base`), so first-call download overhead was not observed in this session — initial users without the cache should expect ~10–30 s on the first `listen()` while the ~140 MB base model downloads

## Trials

Four trials with `duration_ms=5000`, `language="ja"`:

| Trial | Condition | text | frame_count | duration_ms |
|---|---|---|---|---|
| 1 | Distant speech (~30 cm) | `""` | 85 | 5100 |
| 2 | Close-range, short utterance | `"テストテスト コニチュアテストコニチュアテスト"` | 80 | 4800 |
| 3 | Close-range, short utterance | `"マイクのケストレース"` | 79 | 4740 |
| 4 | Complete silence (5 s capture) | `""` | 78 | 4680 |

## Observations

- **Wire path is solid.** Inbound Opus frames arrive at the expected ~60 ms cadence (78–85 frames per 5 s window). Decode and transcription complete within the capture window plus a small overhead.
- **Recognition accuracy is bounded by the default `base` model + `beam_size=1`.** Short Japanese utterances are recognised but show consonant confusion ("テ" → "ケ") and trailing hallucination on dead air after a short phrase. Longer utterances or `model="medium"` are likely to improve accuracy; this is configurable per call.
- **Capture-side gain looks under-set for the CoreS3 mic.** Distant speech (Trial 1) produced 85 frames of audio but zero transcription. Close-range speech (Trials 2–3) transcribed reliably. The CoreS3 `input_gain_` is currently `30` dB in `cores3_audio_codec.cc`, while other boards in this firmware (`esp-box-lite`, `df-k10`) explicitly raise theirs to `37.5` dB with a comment noting the same "mic too quiet" problem. A follow-up to bump CoreS3 to 37.5 dB looks worth filing.
- **Silence is safe.** Complete-silence captures (Trial 4) returned an empty string with no hallucinated text, even with `vad_filter=False`. Faint-signal-plus-noise (Trials 2–3) is where hallucination shows up.
- **Tailscale Funnel ~9 s disconnect (#64) was not observed.** Each `listen()` round-trip completed inside ~5 s, well under the threshold. Longer captures (e.g. `duration_ms=15000`) are not covered by this session.

## Follow-up candidates

- Raise CoreS3 `input_gain_` from 30 to 37.5 dB to match the precedent set by `esp-box-lite` / `df-k10` (small firmware-only change).
- Visual feedback during `listen()`: surface the listening state on the device avatar automatically so the caller can align their utterance to the capture window. Currently this is done manually via `set_avatar("thinking")` before `listen()`.

Issue #115, opened 2026-05-14
feat(firmware): add boot-time servo initialization to a fall-safe neutral pose (yaw=0, pitch=45)


kisaragi-mochi:
## Problem

The firmware's `InitializeServo()` currently only opens the UART bus and reads `current_deg` from `ReadPos()` — **it does not actively position the head at boot**. This has two practical consequences:

1. **Fall-risk pose retention.** Whichever angle the head was left at when power was removed (including hand-pushed positions or end-stop contact) is the angle the device boots into. There is no defined "safe rest pose".
2. **Cold-boot bus-hang trigger.** If power-off happened with pitch near `0°` (mechanical end-stop region), the first `move_head` call into the M5Stack-recommended `5..85°` range can trigger the SCS0009 bus hang documented in #100 — `get_head_angles` then returns `{"yaw":-144,"pitch":-194}` (the `ReadPos == -1` sentinel) and only a CoreS3 power-button long-press recovers the device. Real-device evidence in #100 confirms `pitch=0° → pitch=5°` (a `+5°` motion, the smallest move into the recommended floor) is enough to reproduce.

Confirmed by inspecting both the local firmware (`firmware/main/boards/stackchan/stackchan.cc` `InitializeServo()`) and its direct upstream `kisaragi-mochi/xiaozhi-esp32` `main/boards/stackchan/stackchan.cc` — neither path issues a boot-time `WritePos`.

## Goal

Boot every session at a deterministic, fall-safe neutral pose so that:

- The head leaves boot in a known angle, not in whatever pose the previous session ended in.
- The first user-driven motion always starts from inside the M5Stack-recommended `5..85°` pitch range, eliminating the cold-boot bus-hang surface for #100.

Proposed neutral pose: **`yaw=0°, pitch=45°`** — the center of the M5Stack-recommended `5..85°` range, maximizing safety margin from both end-stops. This is also close to the `look_up_pitch=50°` default landed in PR #107 and aligns with the natural "head-level / slight upward gaze" pose.

## Reference implementations (prefer reuse over reinvention)

| Source | Pattern | Notes |
|---|---|---|
[...]

Issue #110, opened 2026-05-14
ws:// LAN gateway URL ignored — firmware falls back to compiled-in wss://api.tenclass.net default


tluxton:
**Environment**
- Hardware: M5Stack StackChan (Kickstarter 2025 kit, CoreS3)
- Firmware: firmware-v1.3.0 (merged-binary.bin)
- Gateway: stackchan-mcp 0.6.0
- Pi 4 8GB, self-hosted gateway on LAN

**What I did**
Flashed firmware-v1.3.0 clean (merged-binary.bin to 0x0).
Provisioned WiFi via 192.168.4.1.
Set WebSocket Gateway URL to: ws://192.168.0.208:8765

Verified via esptool NVS readback that the value is correctly 
written to the websocket namespace, url key:

  websocket
  nurl
  ws://192.168.0.208:8765

Partition table confirms NVS at 0x9000, size 0x4000.
NVS readback is 0x4000 bytes. Write verified.

**What happens**
Serial log on wake word trigger shows:

  WS: Adding websocket gateway candidate from websocket.url: 
  wss://api.tenclass.net/xiaozhi/v1/

The firmware logs it's reading from websocket.url but returns 
the compiled-in default, not the NVS value.

**Question**
Does the firmware validate that the gateway URL starts with 
wss:// and fall back to the Kconfig default if it starts with 
ws://? Is plain ws:// supported for LAN use, or is TLS required?

If wss:// is required, is there a recommended approach for 
self-hosted LAN gateways — e.g. nginx TLS termination, or a 
build flag to allow ws://?

kisaragi-mochi:
Thanks for the detailed report — confirming the NVS contents with `esptool` was exactly the right diagnostic step, and it points at the real cause rather than a scheme filter.

**Short answer:** `ws://` LAN URLs are fully supported. The firmware does not validate the URL scheme, and TLS / nginx termination is not required for self-hosted LAN gateways.

## Root cause

The upstream `xiaozhi-esp32` OTA-config code path is still active in this fork. On every boot, after WiFi connects:

1. `Ota::CheckVersion()` POSTs to `CONFIG_OTA_URL`, which defaults to `https://api.tenclass.net/xiaozhi/ota/`. This default is not overridden in `firmware/sdkconfig.defaults` or in `firmware/main/boards/stackchan/config.json`, so the released `firmware-v1.3.0` binary hits the upstream server.
[...]

PR #120, opened 2026-05-14
docs(changelog): backfill #98 two-tier pitch guard entry under [Unreleased] Firmware

### LINKED ISSUE #98: fix(firmware): adopt two-tier pitch guard (hard clamp vs. M5Stack-recommended range)
## Context

PR #81 (#80) introduced a defense-in-depth pitch clamp at `SAFE_PITCH_MIN = 0` / `SAFE_PITCH_MAX = 30` in `firmware/main/boards/stackchan/stackchan.cc`. The lower bound `0°` was chosen as the validated mechanical end-stop (~-1°) plus a 1° safety margin, while keeping the "neutral = 0°" API convention. That part is correct and stays.

The upper bound of `+30°` was carried over from the original property declaration (`-30..+30`) without explicit hardware justification. It is well below the M5Stack-recommended ceiling of `+85°` and is the bottleneck blocking the look-up variant in #96 and several other gesture-oriented features.

## Proposed design: two-tier guard

The current code has only one layer of pitch protection (`SAFE_PITCH_MIN/MAX`), which conflates two different concerns:

1. **Hard clamp** = the absolute mechanical limit, just inside the validated end-stop. Its only job is to prevent physical damage to the servo. Edging right up to this is allowed; passing it would break hardware.
2. **Recommended operating range** = the M5Stack-documented sweet spot for long-term servo reliability. Inside this window the servo is comfortable; just outside it but inside the hard clamp may be fine occasionally but is not endorsed.

PR #81 implicitly chose tier (1) for the lower bound (0° is the mechanical-margin floor, not the M5Stack 5° recommended floor) without naming the distinction. We should make the two tiers explicit and apply the same logic to the upper bound.

### Current state (before this fix)

| Layer | Lower | Upper |
|---|---|---|
| Mechanical limit (validated) | ~-1° (PR #79) | **TBD** — not yet validated, likely ~90° per SCS0009 + StackChan kit geometry |
| Firmware hard clamp (`SAFE_PITCH_*`) | 0° | **30°** ← bottleneck |
| MCP tool description / README | "recommended 0..30°" | (same) |
| M5Stack official recommendation | 5° | 85° |

### Proposed state (after this fix)

| Layer | Lower | Upper |
|---|---|---|
| Mechanical limit | ~-1° (already validated) | **TBD — validate on real device** |
| Firmware hard clamp (`SAFE_PITCH_*`) | **0°** (unchanged, +1° margin from validated mechanical end-stop) | **NEW_MAX** (1° inside the newly-validated mechanical upper limit, likely 89° or 90°) |
| MCP tool description / README | **"recommended 5..85° per M5Stack docs"** (both bounds) | (same) |

Both tiers get communicated:

- The firmware refuses physically dangerous values (hard clamp).
- LLM agents and human users see the M5Stack-recommended `5..85°` in the tool description / README and stay there by default.
- Edge-of-range values (e.g. `pitch=87°`) inside the hard clamp but outside the recommended range are accepted with a softer signal (e.g. `ESP_LOGI` or no log) rather than the current hard rejection — they are not hardware-damaging.

## Why keep `SAFE_PITCH_MIN = 0` rather than tightening to 5°

PR #81 already encoded this trade-off; the two-tier framing now makes it explicit:

- The validated mechanical end-stop sits at ~-1°, **not** ~5°. The hard clamp's job is to prevent mechanical damage; it succeeds at `0°` (+1° margin from the validated end-stop).
- `pitch = 0` is the API's "neutral / horizontal" convention. Tightening the hard clamp to `5°` would invalidate the natural neutral, breaking existing callers and making `set_head_angles(pitch=0)` log a clamp warning every time.
- M5Stack's `5°` lives in the **description / README layer**, not the firmware hard clamp. Agents are nudged to stay above 5° via documentation rather than blocked at the firmware level.

## Proposed change

In `firmware/main/boards/stackchan/stackchan.cc`:

- `SAFE_PITCH_MIN` stays at `0` (unchanged).
- `SAFE_PITCH_MAX`: `30` → **TBD** (1° inside the newly-validated mechanical upper limit; pick during real-device validation, likely `89` or `90`).
- `set_head_angles` MCP property declaration: `Property("pitch", kPropertyTypeInteger, 0, -30, 30)` → `Property("pitch", kPropertyTypeInteger, 0, -30, NEW_MAX)`.
- `set_head_angles` MCP tool description string: surface the **M5Stack-recommended range** prominently while noting that the firmware accepts up to `NEW_MAX°`:
  > "yaw: horizontal (-90 to 90). pitch: vertical, M5Stack-recommended operating range 5 to 85 per https://docs.m5stack.com/en/StackChan. The firmware also accepts values up to NEW_MAX° (hard clamp guards against mechanical end-stop damage), but values outside 5-85° are not officially endorsed and may stress the servo over time."
- `README.md` / `README.ja.md` "Hardware safety notes" section (EN/JP both): rewrite to document the two-tier model explicitly — the M5Stack-recommended `5..85°` as the operational sweet spot, the firmware hard clamp at `0..NEW_MAX°` as the safety net for occasional edge-of-range use.

## Validation

The mechanical upper limit needs to be determined on a real device before `SAFE_PITCH_MAX` can be finalised:

- Flash test firmware with a temporarily-raised clamp (e.g. `SAFE_PITCH_MAX = 95`).
- Run `set_head_angles(yaw=0, pitch=N)` for `N` in ascending `{85, 87, 89, 90, 92}`; inspect actual `ReadPos`, listen for audible end-stop hit, watch for servo gear strain or stall.
- Pick `SAFE_PITCH_MAX` = (validated upper end-stop) - 1°.
- Confirm `set_head_angles(pitch=85)` operates comfortably inside the recommended ceiling.
- Sweep `set_head_angles(pitch=N)` for `N` in `{0, 5, 30, 45, 60, 85, NEW_MAX}` and visually verify smooth head motion through the full range; check motion-task interpolation does not overshoot or oscillate.

## Why this matters

- **Issue #96** (visual + motion feedback during `listen()`) wants a "look-up" pose around 45°–60° pitch — blocked by the current `30°` hard clamp.
- **Issue #89** (named motion templates) would naturally include "look up at the ceiling" / "look up curiously" expressions — mostly blocked above 30°.
- Future affection / acknowledgement / glance gestures from PR #4 / #6 (touch reactions) would benefit from the full upward range.
- The two-tier model also documents the lower-bound rationale (PR #81) which was previously implicit.

## Risks

- The new `SAFE_PITCH_MAX` is a defense-in-depth path — pitch in `(30..NEW_MAX]` was silently clamped before and would be accepted afterward. Real-device validation gates the final value.
- The MCP property declaration changes from `+30` to `NEW_MAX`. This is **backward-compatible** for existing callers (requests ≤30° still work); clients pinned to the old maximum will simply not exercise the new range.
- The softer treatment of `5°` lower / `85°` upper recommended boundaries (description-only, no hard rejection) shifts responsibility for "stay inside the M5Stack recommended range" to callers / LLM prompts. Acceptable trade-off because (a) hard clamp still prevents damage and (b) hard-blocking edge-of-recommended values would surprise agents that hit `pitch=86` once in a while.

## Out of scope

- Yaw (X-axis): M5Stack docs state no restriction. No change to `YawDegToPos` / declared `-90..+90°` range.
- Pitch lower hard clamp: stays at `0°` for the reasons above.

Refs #80, #81 (lower-bound clamp — this Issue makes its two-tier intent explicit).
Refs #96 (depends on the upper-bound widening for the look-up variant), #89, #4, #6.

kisaragi-mochi:
## Summary

PR #101 (closes #98) was merged on 2026-05-13 after firmware-v1.3.0 was
tagged, but did not add an entry to `CHANGELOG.md`. The
firmware-v1.4.0 release notes are auto-extracted from the
`[Unreleased]` Firmware section, so without this backfill the release
notes would only mention #115 and #110 and silently omit the
user-visible change to `set_head_angles` pitch handling.

This PR adds the missing entry verbatim from the PR #101 commit
message and the README "Hardware safety notes" rewrite, so the
firmware-v1.4.0 release cut on the same day can pick it up cleanly.

## Changes

- `CHANGELOG.md`: add a `### Firmware` bullet under `[Unreleased]`
  describing the two-tier pitch guard (`0..88` hard clamp / `5..85`
  recommended range) on `set_head_angles`.

Documentation-only — no code change. No other files touched.

## Test plan

- [ ] Render the `[Unreleased]` block locally and confirm the three
      Firmware bullets (#98 / #115 / #110) appear in the intended
      order for firmware-v1.4.0 release notes extraction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,22 @@ change is called out under a `Firmware` subsection of the release entry.
 
 ### Firmware
 
+- Replaced the single-bound pitch clamp (`0..30°`) on `set_head_angles`
+  with a two-tier guard: a hard mechanical clamp at `0..88°` (Tier 1,
+  always enforced with `ESP_LOGW` on clip) and an `ESP_LOGI` advisory
+  for requests outside the M5Stack-documented operating range `5..85°`
+  (Tier 2). This matches the gateway-side `move_head` policy
+  introduced in [0.7.0] (gateway tool enforces the restrictive `5..85`
+  range as the authoritative Tier 1 boundary at the MCP layer, while
+  the firmware retains the permissive `0..88` hardware-safety clamp so
+  `set_head_angles` callers that genuinely need the wider range are
+  still served). README EN/JP "Hardware safety notes" rewritten as a
+  two-tier table. Validated on M5Stack CoreS3 + SCS0009 ×2. Closes
+  [#98](https://github.com/kisaragi-mochi/stackchan-mcp/issues/98).
+  Refs
+  [#80](https://github.com/kisaragi-mochi/stackchan-mcp/issues/80),
+  [#81](https://github.com/kisaragi-mochi/stackchan-mcp/issues/81).
+
 - Fixed user-configured WebSocket gateway URLs (e.g.
   `ws://192.168.x.y:8765`) being silently overwritten on every boot by
   the upstream xiaozhi OTA-config response. `Ota::CheckVersion()` still


kisaragi-mochi:
Closing — proceeding with the firmware-v1.4.0 release from current main as-is. The #98 entry can be folded into a later Gateway release's Firmware subsection along with the other [Unreleased] entries.

PR #136, opened 2026-05-15
fix: keep WebSocket alive when CloseAudioChannel is called

Qizhan7:
## Problem

`CloseAudioChannel()` in `websocket_protocol.cc` calls `websocket_.reset()`, which destroys the WebSocket connection every time the device exits listening/speaking mode. This makes it impossible to use MCP tools (LEDs, avatar, head movement) outside of an active audio session — all control is lost the moment a conversation ends.

The current workaround in the community is to never send `tts.stop`, keeping the device in speaking state. This works but is a hack that prevents proper state management.

## Root Cause

```cpp
void WebsocketProtocol::CloseAudioChannel(bool send_goodbye) {
    ...
    websocket_.reset();  // destroys the connection
}
```

The call chain `CloseAudioChannel() → websocket_.reset() → OnDisconnected → on_audio_channel_closed_()` couples the audio session lifecycle to the WebSocket connection lifecycle. These should be independent — MCP control needs to work regardless of audio state.

## Fix

Skip the WebSocket teardown and directly invoke `on_audio_channel_closed_()`:

```cpp
void WebsocketProtocol::CloseAudioChannel(bool send_goodbye) {
    (void)send_goodbye;
    ESP_LOGI(TAG, "CloseAudioChannel: keeping WebSocket alive for MCP");
    if (on_audio_channel_closed_ != nullptr) {
        on_audio_channel_closed_();
    }
}
```

The app transitions back to idle correctly. The WebSocket stays connected for MCP. If the connection drops due to network issues, `OnDisconnected` still fires normally and `ScheduleReconnect` handles recovery.

## Testing

- Flash, reboot, WebSocket connects to gateway
- Touch screen to enter listening, confirm WebSocket stays connected
- Touch again to exit listening, confirm WebSocket still alive
- Call MCP tools (set_avatar, move_head, set_all_leds), all work in idle state

Tested on M5Stack CoreS3 with stackchan board config.

### DIFF
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -383,6 +383,13 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error) {
 
         websocket_->OnDisconnected([this, notify_disconnect]() {
             audio_channel_open_.store(false);
+            // notify_disconnect carries this socket's reconnect intent.
+            // ParseServerHello() arms it (true) once the handshake
+            // completes; intentional teardown paths (CloseAudioChannel,
+            // OpenAudioChannelInternal, destructor) disarm it (false)
+            // before resetting the socket. A false reading here means
+            // either the candidate never completed handshake or the
+            // close was intentional — neither should reconnect.
             if (!notify_disconnect->load()) {
                 ESP_LOGI(TAG, "Websocket disconnected (no reconnect: candidate failed or intentional close)");
                 return;
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -147,18 +147,23 @@ bool WebsocketProtocol::IsAudioChannelOpened() const {
 }
 
 void WebsocketProtocol::CloseAudioChannel(bool send_goodbye) {
-    (void)send_goodbye;  // Websocket doesn't need to send goodbye message
-    // Mark the close as intentional so any reconnect job already
-    // re-posted from the timer callback aborts when it runs on the main
-    // task, then disarm the current socket's per-socket flag so the
-    // OnDisconnected lambda hits the early-return guard the moment the
-    // underlying close fires (the lambda runs on the WS task).
-    intentional_close_.store(true);
-    if (current_notify_disconnect_) {
-        current_notify_disconnect_->store(false);
+    (void)send_goodbye;
+    // Keep WebSocket alive — only notify the application that the audio
+    // channel is logically closed so it returns to idle state.
+    //
+    // The original implementation called websocket_.reset() here, which
+    // destroyed the WebSocket connection every time the device exited
+    // listening/speaking mode. This made it impossible to control the
+    // device (LEDs, avatar, head movement) outside of an active audio
+    // session, since all MCP tools rely on the same WebSocket.
+    //
+    // By skipping the teardown and directly invoking the closed callback,
+    // the app transitions back to idle while the WebSocket stays connected
+    // for continued MCP control.
+    ESP_LOGI(TAG, "CloseAudioChannel: keeping WebSocket alive for MCP");
+    if (on_audio_channel_closed_ != nullptr) {
+        on_audio_channel_closed_();
     }
-    StopReconnectTimer();
-    websocket_.reset();
 }
 
 bool WebsocketProtocol::OpenAudioChannel() {
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -60,7 +60,7 @@ WebsocketProtocol::WebsocketProtocol() {
                 }
 
                 ESP_LOGI(TAG, "Reconnecting to websocket server");
-                if (!protocol->OpenAudioChannelInternal(false)) {
+                if (!protocol->OpenAudioChannelInternal(false, false)) {
                     ESP_LOGW(TAG, "Reconnect attempt failed; rescheduling");
                     protocol->ScheduleReconnect();
                 }
@@ -168,10 +168,10 @@ void WebsocketProtocol::CloseAudioChannel(bool send_goodbye) {
 }
 
 bool WebsocketProtocol::OpenAudioChannel() {
-    return OpenAudioChannelInternal(true);
+    return OpenAudioChannelInternal(true, true);
 }
 
-bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error) {
+bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_audio_channel) {
     // Resetting the previous websocket may invoke its OnDisconnected
     // callback synchronously. Disarm the previous socket's flag and
     // mark the teardown as intentional so neither the per-socket lambda
@@ -311,7 +311,7 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error) {
         websocket_->SetHeader("Device-Id", SystemInfo::GetMacAddress().c_str());
         websocket_->SetHeader("Client-Id", Board::GetInstance().GetUuid().c_str());
 
-        websocket_->OnData([this, notify_disconnect](const char* data, size_t len, bool binary) {
+        websocket_->OnData([this, notify_disconnect, arm_audio_channel](const char* data, size_t len, bool binary) {
             if (binary) {
                 // Drop inbound audio when the audio channel is logically
                 // closed. Without this guard, a late TTS frame from the
@@ -359,7 +359,7 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error) {
                 auto type = cJSON_GetObjectItem(root, "type");
                 if (cJSON_IsString(type)) {
                     if (strcmp(type->valuestring, "hello") == 0) {
-                        ParseServerHello(root, notify_disconnect);
+                        ParseServerHello(root, notify_disconnect, arm_audio_channel);
                     } else if (!audio_channel_open_.load() &&
                                (strcmp(type->valuestring, "tts") == 0 ||
                                 strcmp(type->valuestring, "listen") == 0)) {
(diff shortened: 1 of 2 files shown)


kisaragi-mochi (commented):
Thanks for opening this PR! The problem you're describing — losing MCP control whenever the audio session ends — is one we've also hit, and the direction of "keep the WebSocket alive across audio-session lifecycle" is exactly the right framing.

Before merge I ran an adversarial codex review on the diff with focus on reconnect-logic and lifecycle interactions. It surfaced two concerns worth raising. The core fix direction is the right shape; these are about completing the separation between "audio channel state" and "control WebSocket state".

## Finding 1 (high): `CloseAudioChannel` leaves `IsAudioChannelOpened()` returning `true`

In the current diff, `CloseAudioChannel()` only invokes `on_audio_channel_closed_()`. The audio-channel-open state is implicitly inferred from `websocket_->IsConnected()` (see `IsAudioChannelOpened()` in the same file), so as long as the WS stays alive, callers see the channel as still open.

Side effects:

- `Application::StartListening()` / `ToggleChatState()` check `IsAudioChannelOpened()` and skip `OpenAudioChannel()` when it reports `true`. That means the next audio session does **not** reset `session_id_`, does **not** send a fresh `hello`, and does **not** invoke `on_audio_channel_opened_()`. Subsequent sessions can replay against the previous server-side session id, which the gateway tracks for state transitions.
- `Application::CanEnterSleepMode()` treats `IsAudioChannelOpened() == true` as an active audio channel, so the device cannot enter sleep mode once an audio session has been opened, even when it is now logically idle.

Suggested approach: split the logical audio-channel state from the physical WebSocket state. For example, add a dedicated `audio_channel_open_` member set to `true` after a successful `OpenAudioChannel()` + hello exchange, and to `false` in `CloseAudioChannel()` / `OnDisconnected()`. `IsAudioChannelOpened()` then returns `audio_channel_open_` instead of `websocket_->IsConnected()`. The MCP layer can use a separate predicate (e.g. `IsControlChannelAvailable() → websocket_ && websocket_->IsConnected()`) for its own check.

## Finding 2 (high): Stale audio-session WS messages can still resurrect the audio workflow

Without an explicit logical-close marker, late or already-in-flight server messages from the just-closed session still flow into `Application::OnIncomingJson()`. Concretely: a `tts.start` arriving after `CloseAudioChannel()` would schedule `kDeviceStateSpeaking` even though the local close callback has just scheduled `kDeviceStateIdle`.

Previously the `websocket_.reset()` was the implicit barrier that ensured no further audio-session traffic could arrive once the local close fired.

Suggested approach: after `CloseAudioChannel()`, if the WebSocket is kept alive, either (a) skip dispatch of audio-session JSON message types (`tts.*`, `listen.*`, binary audio frames) until a fresh `OpenAudioChannel()` re-arms the session, or (b) check the incoming `session_id` against the currently-armed session id and drop mismatches. Pattern (b) integrates naturally with Finding 1's `audio_channel_open_` flag — drop messages whose session id does not match while the flag is `false`.

## Removed reconnect-logic concerns (for completeness)

The codex review also examined the four pieces removed from the original `CloseAudioChannel()`:

- `intentional_close_.store(true)` (from #35)
- `current_notify_disconnect_->store(false)` (from #35)
- `StopReconnectTimer()`
- `websocket_.reset()`

These are not flagged as immediate regressions because `CloseAudioChannel()` is no longer tearing down the socket. The `intentional_close_` flag was specifically the close-versus-network-disconnect signal for the `OnDisconnected` lambda; if the socket is never reset on `CloseAudioChannel()`, the lambda only fires from a true network disconnect, where reconnect-on-disconnect is the desired path. That part is internally consistent.

However: if any future code path explicitly tears down the WebSocket (e.g. an explicit `Disconnect` MCP tool, device reboot, or `OpenAudioChannel()` after a network drop), the `audio_channel_open_` flag from Finding 1 should also reset to `false` at that point so the state remains consistent.

## Testing suggestions

A minimal coverage list to demonstrate the corner cases on real hardware:

1. After audio session ends + `CloseAudioChannel()`: call MCP tools (LED, avatar, `move_head`). Confirm they work (this PR's stated benefit).
2. Same scenario: trigger touch / wake-word to start a new audio session. Confirm `OpenAudioChannel()` is reached, `session_id_` is fresh, `hello` is sent.
3. Inject a delayed `tts.start` from the server immediately after `CloseAudioChannel()`. Confirm the device does not re-enter `kDeviceStateSpeaking`.
4. After audio session ends: confirm `CanEnterSleepMode()` allows the device to enter power-save mode (currently it would not, per Finding 1).

## Repository owner sign-off

Leaving this as a code review comment rather than an approval, since the architectural choice (control-WS vs audio-channel split) deserves repo-owner sign-off. Once the two findings above are addressed, the underlying fix direction is solid and the testing you've already done on M5Stack CoreS3 confirms the happy path works.

Happy to iterate on the patch if helpful. Thanks again for the contribution.

kisaragi-mochi (commented):
Thanks for the quick turnaround on the two findings — `audio_channel_open_` as a separate logical state and the symmetric drop logic in `OnData` for binary frames and `tts`/`listen` JSON look right and resolve the original concerns.

Before merge I ran another adversarial codex pass focused on reconnect-path interactions and it surfaced one further concern. This one is reconnect-specific and only becomes observable now that `CloseAudioChannel()` keeps the WebSocket alive.

## Finding 3 (high): Auto-reconnect re-arms a logically-closed audio channel

Path:

1. User triggers `CloseAudioChannel()` → `audio_channel_open_ = false`, WebSocket stays alive.
2. While the WebSocket is idle, the transport drops (network blip, gateway restart, Tailscale Funnel TLS reset, etc.).
3. `WebsocketProtocol::OnDisconnected` callback fires. `notify_disconnect` is `true` (armed back when the original `ParseServerHello` ran), so the lambda invokes `on_disconnected_()`, `on_audio_channel_closed_()`, and `ScheduleReconnect()`.
4. The reconnect timer calls `OpenAudioChannelInternal(false)` (`websocket_protocol.cc:63`).
5. New socket + new hello → server replies with hello → `ParseServerHello()` runs → `audio_channel_open_.store(true)` at `websocket_protocol.cc:587`.

At step 5 the logical audio channel is open again **even though the user closed it in step 1**. The Finding 1 / Finding 2 protections then no longer apply:

- `IsAudioChannelOpened()` returns `true`
- `CanEnterSleepMode()` blocks sleep
- Binary-frame and `tts.*` / `listen.*` JSON drop guards in `OnData` are disabled
- A late `tts.start` from the server (or a new one sent in response to the new hello) schedules `kDeviceStateSpeaking` against the user's intent

This is essentially Finding 2 reappearing through the reconnect door.

### Suggested approach

Split "WebSocket transport up" from "audio session armed" along the existing user-vs-reconnect boundary you already have between `OpenAudioChannel()` (user) and `OpenAudioChannelInternal(false)` (reconnect timer).

Concretely:

- Thread the intent through to `ParseServerHello`. For example:
  ```cpp
  bool OpenAudioChannelInternal(bool report_error, bool arm_audio_channel);
  void ParseServerHello(const cJSON* root,
                        const std::shared_ptr<std::atomic<bool>>& notify_disconnect,
                        bool arm_audio_channel);
  ```
- `OpenAudioChannel()` (user) → `OpenAudioChannelInternal(true, true)`.
- Reconnect timer path (`websocket_protocol.cc:63`) → `OpenAudioChannelInternal(false, false)`.
- In `ParseServerHello`, set `audio_channel_open_.store(arm_audio_channel)` instead of unconditional `true`.
- Consider whether `on_audio_channel_opened_()` (`websocket_protocol.cc:450-452`) should also be gated on `arm_audio_channel` — for a pure transport reconnect, firing the audio-channel-open callback is arguably also out of intent. Happy to defer that to a follow-up if it's outside the scope of this PR.

### Minor observation (separate, not a blocker)

`CloseAudioChannel(bool send_goodbye)` currently discards `send_goodbye` via `(void)send_goodbye;`. The pre-PR implementation may have actually sent a goodbye to the server to signal session end; without that, the server may keep the session live and stream a late `tts.*` toward the client. The Finding 2 drop logic handles this on the client side, but server-side cleanup may also matter for the gateway. Worth a follow-up regardless of how Finding 3 is resolved.

### Testing addition

To the four cases from the previous review, please add:

5. After `CloseAudioChannel()`, simulate or wait for a transport drop (e.g. restart the gateway). Confirm that once reconnect completes, `IsAudioChannelOpened()` still returns `false`, `CanEnterSleepMode()` still permits sleep, and a server-injected late `tts.start` does not transition the device to `kDeviceStateSpeaking`.

### Sign-off

The original direction is solid, and Finding 1 / Finding 2 are addressed correctly. Finding 3 is the same lifecycle question one layer deeper — it only becomes observable now that the socket survives `CloseAudioChannel()`. Happy to keep iterating; thanks again for the contribution.

kisaragi-mochi:
Thank you for the thorough iteration on this — three review rounds (Findings 1+2, Finding 3) have shaped a much cleaner separation between the WebSocket transport lifetime and the audio-session lifetime. The original framing (keep the WS alive across `CloseAudioChannel`) was right, and the layered handling of stale frames + the `arm_audio_channel` boundary closes the lifecycle questions raised in review.

Before merging I ran one further adversarial codex pass to look for follow-on effects. It surfaced three additional observations, but **none of them require changes to this PR** — they are tracked separately so the work you've already done lands as-is:

- **Finding A** — race in `OnDisconnected` `audio_channel_open_` write ordering (low real-world probability). Tracked as #188.
- **Finding B** — hello-then-disconnect race in `OpenAudioChannelInternal` can cancel the reconnect timer (low real-world probability). Tracked as #189.
- **Finding C** — stackchan-mcp-specific: the gateway sends gateway-driven `tts.start` / `listen.start` MCP messages to drive the firmware state machine (`gateway/stackchan_mcp/esp32_client.py:222-268`), which the new drop guard would silently discard after a user-initiated `CloseAudioChannel()`. This is not visible from the firmware-side perspective alone. Tracked as #187 with a companion fix from the repo side using session_id-based gating, which preserves your stale-frame protection while letting current-session messages flow through.

Merging now. Thanks again for the careful work and patience through the rounds.

kisaragi-mochi:
Quick follow-up: companion fix for the Finding C call-out in the merge comment has landed as #192 (currently in review on the repo side). No action needed from this PR — the original `websocket_.reset()` deletion + `arm_audio_channel` lifecycle work you contributed here is preserved as-is; #192 only changes the JSON drop predicate from `audio_channel_open_` to a `session_id` match, so the gateway-driven `tts.start` / `listen.start` path that this repo's MCP tools rely on flows through cleanly.

Thanks again for the careful review-round work on this.

Issue #140, opened 2026-05-16
feat(firmware): hardware-level boot snap-motion prevention via PY32-driven VM_EN gradual ramp


kisaragi-mochi:
## Background

PR #137 (snap-suppress hold) and PR #139 / Issue #138 (`ReadPos` retry + safe-fallback seed) provide firmware-side mitigation for the boot-time snap-motion documented in #121 Problem 1. Both intercept the motion *after* it begins, by either writing the current position as a new servo set-point (#137) or by keeping the firmware-side interpolation a no-op (#138). They do not prevent the underlying SCS0009 self-initiated motion: the servo always tries to slew toward its last commanded set-point as soon as `VM_EN` HIGH restores torque.

A more fundamental fix would prevent the snap motion from starting in the first place. The most plausible avenue is hardware-level: the PY32 IO expander already controls `VM_EN`, currently driven instantaneously from LOW to HIGH at boot (`Servo power ENABLED via PY32 pin 0 (VM EN HIGH confirmed)` log line). Gradually ramping `VM_EN` HIGH — or staying at a reduced level for a configurable window — would let the servo wake with reduced torque, so any self-initiated motion cannot reach the end-stop trajectory that produces the "thunk" + bus-hang failure mode observed in #121 / #100.

## Investigation / proposed work

- Confirm with the PY32 firmware (versioned `0x41` per the current boot log) whether gradual / staged `VM_EN` control is possible. Likely requires extending the PY32 register set or adding a new IO-expander command.
- Characterize how much torque attenuation is needed to suppress the snap-motion trajectory without breaking the boot-init `WriteHeadAngles` climb to `(yaw=0°, pitch=45°)` over 4 seconds.
- Consider partial alternatives at the firmware level (e.g. a SCS0009 torque-disable command if one exists, issued before the first `WritePos`, then re-enabled after `BOOT_INIT_PITCH_DEG` is reached).
- Compare measured snap-motion amplitude with and without the ramp, using the existing `Boot pre-init ReadPos` diagnostic line (raw position + attempt count) plus the post-init `ReadPos` already logged at the end of `InitializeServo()`.

## Acceptance

This Issue is intentionally tracker-style. Goal is to keep a public record of the root-cause direction so that a future contributor (or the same one, when the PY32 firmware side is revisited) can resume from where this thread left off.

Closing condition is one of:

[...]

Issue #138, opened 2026-05-15
fix(firmware): ReadPos retry + safe current_deg fallback when boot pre-init reads land in SCS0009 startup-latency window


kisaragi-mochi:
## Background

PR #137 added a boot-time snap-suppress hold `WritePos` in `InitializeServo()` as a partial mitigation for #121 Problem 1 (downward drop on power-on). The hold catches snap motion when the SCS0009 bus is responsive at boot time, but is intentionally skipped when `Boot pre-init ReadPos` returns `-1`, so that unknown / out-of-range positions are never written back.

Real-device verification during PR #137 surfaced two observations that motivate this follow-up:

### 1. SCS0009 has a `~200 ms` startup latency after `VM_EN` HIGH

During real-device verification of PR #137 on the PMIC long-press OFF / ON path (which cycles `VM_EN` LOW → HIGH and forces the SCS0009 to wake up from a fully powered-down state), both `ReadPos` and `WritePos` returned `-1` for approximately the first `~200 ms` after `Servo power ENABLED via PY32` logged. This is distinct from the SCS0009 bus hang tracked in #100 (which requires a PMIC long-press to clear). The startup latency self-clears as the servo wakes up; subsequent `WritePos` succeed and the bus is fully responsive thereafter.

The current `Boot pre-init ReadPos` runs at tick ~140 ms, **inside** that latency window, so both reads typically return `-1` on the PMIC OFF/ON path even when the servo is healthy.

### 2. When `ReadPos` fails at boot, the firmware-side `current_deg` fallback is currently `0`

`MotionState::current_deg` defaults to `0` at construction. When `Boot pre-init ReadPos` returns `-1`, `InitializeServo()` logs `Failed to ReadPos(pitch); current_deg stays at 0` and leaves `pitch_motion_.current_deg` at that default.

`pitch_motion_.current_deg=0` corresponds to `pos=620` — the lower mechanical end-stop, **outside the M5Stack-recommended `5..85°` operating range**. The subsequent boot-init `WriteHeadAngles(0, BOOT_INIT_PITCH_DEG=45, 4000ms)` interpolation then walks the `WritePos` sequence from `deg=0` upward (`pos=620, 623, 626, 629, ...` at 50 ms intervals from the `ServoTask` interpolator), issuing repeated `WritePos` calls at and just above the lower end-stop.

Concrete log from PR #137 verification (PMIC OFF/ON path):

```
[...]

Issue #121, opened 2026-05-14
bug(firmware): boot-time servo motion is unsafe — unintended downward drop on power-on + over-fast 45° climb


kisaragi-mochi:
## Surfaced during real-device verification of PR #117 (Issue #118)

Real-device verification of PR #117's `fall-safe boot-time servo init` revealed two related but distinct firmware-side issues that together undermine the `fall-safe` framing of PR #117. Both observed on M5Stack CoreS3 + SCS0009 ×2 with `CONFIG_STACKCHAN_SERVO_FEETECH=y` (default MIT driver).

### Observation

- **Power OFF**: head stays put (servo torque off, the head is held by static friction and the head's center of gravity).
- **Power ON**: head **actively drops to the lower end-stop with an audible "thud" / "ごんっ" impact** before `InitializeServo()`'s `WriteHeadAngles(0, 45, 1000)` can take effect.
- After the drop, the boot init's 1-second climb to 45° proceeds and reaches the target.

The drop is not free fall — it is a deliberate, fast servo motion with end-stop impact. Free-fall would happen at power-off (when torque is removed), not at power-on. The drop happens *after* torque is restored, which means a positive set point is being commanded somewhere in the boot path.

### Why this is unsafe

- Repeated end-stop impacts will damage the SCS0009 gear train over time.
- This is the exact `pitch=0° → ascending motion` condition that triggers the SCS0009 bus hang documented in #100. PR #117's intent was to avoid that condition; the unintended drop puts every boot back into it.
- It contradicts the `fall-safe neutral pose` framing of PR #117.

### Root cause hypotheses (to investigate via serial monitor capture)

1. **SCS0009 NVS-stored set point**: the servo may auto-restore to the last `WritePos` value on power-up. If a previous session left the set point at `pitch=0°` or below, every subsequent power-on reproduces the dive regardless of any firmware logic.
2. **`setServoPowerEnabled(true)` (PY32 IO expander) torque-restore default position**: the moment the servo bus receives power via the PY32-controlled VM_EN GPIO, the servo may snap to a hardware default position (factory `pos=512` ≈ `pitch=−34°`, clamped at end-stop in our coordinate system).
[...]

Issue #100, opened 2026-05-12
fix(firmware): SCS0009 bus hang on large-angle pitch reversals — ESP32 reset insufficient for recovery


kisaragi-mochi:
## Summary

The SCS0009 servo bus enters an unresponsive state after certain large-angle abrupt pitch reversals (e.g. `pitch=88° → pitch=5°`, a sub-90° reversal that nonetheless exceeds the bus's tolerance for back-to-back frames). Once in this state, `ReadPos` returns `-1` indefinitely while `WritePos` continues to ACK in firmware logs (the firmware-side `uart_write_bytes` still completes, but no half-duplex response frames return from the servo).

ESP32 hard-reset (`esptool.py --before default_reset --after hard_reset chip_id` or pyserial `DTR`/`RTS` toggle) does **not** recover the bus, because the SCS0009 power line is gated by the M5Stack CoreS3's PY32 IO-expander `VM EN` pin (read-only via `check_vm_en`), independent of the ESP32 reset path. The only reliable recovery observed so far is the **CoreS3 hardware power button long-press (≥ 6 s)**, which triggers a full AXP2101 PMIC power-off — only then is `VM EN` released and the servo bus clean on next boot.

## Reproduction

Observed during the on-device sweep / spot check for #98:

1. Power-cycle the CoreS3 to a clean state.
2. `move_head(yaw=0, pitch=89)` — Tier 1 hard clamp engages, motion lands at `pitch=88`.
3. After motion settles (~600 ms), `move_head(yaw=10, pitch=5)` — `pitch=88° → 5° = -83°` reversal.
4. Subsequent `get_head_angles` returns `{"yaw":-144, "pitch":-194}` (i.e. `ReadPos == -1` decoded to degrees).
5. `uart_diag` shows `written=8 / expected=8 / tx_done=ESP_OK` but `buf_after=0` — TX side OK, RX side stuck.
6. ESP32 hard-reset (esptool) re-boots the firmware and reconnects WebSocket, but `ReadPos` still returns `-1` after the new boot completes.
7. CoreS3 power-button long-press → full power-off → button short-press to power on → `ReadPos` returns correct values immediately on the new boot.

Symptom signature (from `uart_diag`):

```json
{"buf_before_status":"ESP_OK","buf_before":0,"written":8,"expected":8,
 "tx_done_status":"ESP_OK","buf_after_status":"ESP_OK","buf_after":0}
```

## Hypothesis

[...]

PR #142, opened 2026-05-16
fix(firmware): #141 post-init ReadPos re-sync to align current_deg with SCS0009 actual position

### LINKED ISSUE #141: fix(firmware): boot-init posture sync between firmware current_deg and SCS0009 actual position after safe-fallback branch
## Background

PR #139 / Issue #138 introduced a safe-fallback seed in `InitializeServo()` Phase 2 that sets `pitch_motion_.current_deg = BOOT_INIT_PITCH_DEG (=45°)` when the pre-init `ReadPos` retries (5×50 ms) all return `-1`. This deliberately avoids the boot-init `WriteHeadAngles(0, 45, 4000)` interpolation walking through end-stop-adjacent positions: without the seed, the interpolator would issue `WritePos` calls from `pos=620` (deg=0, lower mechanical end-stop) upward, risking servo bus degradation if the SCS0009 wakes up mid-sequence.

Real-device verification (PR #139 test plan, hands-off PMIC long-press OFF / ON path) confirmed this prevents the end-stop walk and produces zero physical snap motion: the head did not move at all during boot, including the 4-second `WriteHeadAngles` interpolation window. However, the boot log also shows that the firmware-side `current_deg` (`45°`) ends up several degrees away from the actual servo position. From the PR #139 hands-off run:

```
W (1945) StackChanBoard: Failed to ReadPos(pitch) after 5 attempts; seeded current_deg=45 (BOOT_INIT_PITCH_DEG) to avoid end-stop walk during boot-init climb
I (6055) StackChanBoard: Boot-time servo init complete: target yaw=0 pitch=45 (move=4000ms), post-ReadPos: yaw_raw=460 pitch_raw=744, elapsed_ms=4100
```

`pitch_raw=744 ⇒ ~38.75°` while `current_deg=45°`. The mismatch is **not a fixed value** — it depends entirely on the SCS0009's last commanded set-point before power-off (which the servo retains across power cycles and snaps toward when `VM_EN` HIGH restores torque). Any prior set-point produces a different mismatch magnitude and direction on the next boot.

## Symptom

The mismatch is harmless for normal motion — any non-equal `move_head` target restarts interpolation cleanly — but produces a usability rough edge for clients that issue a target equal to `BOOT_INIT_PITCH_DEG` right after boot:

- `get_head_angles` returns `{yaw:0, pitch:38}` (correct, ReadPos-based actual)
- `move_head(0, 45)` returns `pitch_motion_started: 0` (firmware sees `current_deg == target`, no interpolation starts)
- The head does not move
- From an LLM-side client's perspective: "I asked for 45° but the head didn't move; `get_head_angles` still says 38°." — a contradictory observation that is hard to attribute correctly without firmware-internals knowledge.

## Final implementation (PR pending)

Add a **Phase 0' re-sync step** to `InitializeServo()`, between the existing Phase 0 boot-init climb and the end of `InitializeServo()`:

```
Phase 0  (existing, keep):   WriteHeadAngles(0, BOOT_INIT_PITCH_DEG=45, duration)
                             vTaskDelay(duration + 100)
─── new Phase 0' ────────────────────────────────────────────────────────
Phase 0' (post-init re-sync): retry ReadPos (5×50 ms) for each axis;
                              if successful, unconditionally overwrite
                              firmware-side current_deg / start_deg /
                              target_deg / move_start_ms / moving=false
                              with the actual physical position
```

By the time Phase 0's `vTaskDelay` returns the SCS0009 wake-up latency window is well past, so the Phase 0' ReadPos retry reliably succeeds on the PMIC OFF / ON path that triggered the original mismatch. Subsequent `move_head` calls then interpolate from the real starting position.

### Side change: distance-aware Phase 0 duration

Phase 0's `WriteHeadAngles` duration is now computed from the `current_deg → BOOT_INIT_*_DEG` deltas at a unified angular-speed cap (`BOOT_INIT_TARGET_DEG_PER_SEC = 15 °/s`), floored at `BOOT_INIT_MOVE_MS = 3000 ms` to keep the SCS0009 wake-up latency window covered on the PMIC OFF / ON path. Without this, a pre-power-off `yaw=±90°` set-point would drive boot-init yaw at ~30 °/s on the prior fixed `BOOT_INIT_MOVE_MS=4000 ms`, exceeding the no-stutter Smooth lower bound established under #121 Problem 2.

### License boundary

This change is confined to `firmware/main/boards/stackchan/stackchan.cc` (MIT). No edits to the SCServo_lib files (`SC*.{cc,h}`, GPL-3.0). All existing license headers preserved.

## Acceptance criteria

1. **ESP32-only reset path**: Phase 0 climb to `BOOT_INIT_PITCH_DEG=45°` completes, Phase 0' ReadPos succeeds within the retry budget (logs valid `yaw_raw` / `pitch_raw` values, not `-1`), and `current_deg` ends up matching the post-Phase 0 `ReadPos`-derived actual within `HOMING_VERIFY_TOLERANCE_DEG` precision (typical: 1° or better on a non-degraded bus).
2. **PMIC long-press OFF / ON path (hands-off)**: Phase 0 stays a no-op of effect thanks to the #138 safe-fallback seed; Phase 0' re-syncs `current_deg` with the actual physical position; the post-boot `get_head_angles` matches the firmware-side `current_deg` rather than diverging from it by a prior-set-point-dependent offset.
3. **`move_head` immediately after boot** interpolates from the actual position rather than reporting an immediate no-op when the requested target is `BOOT_INIT_PITCH_DEG`.
4. **No regression on the existing motion paths** (multiple `move → settle → read` cycles work as before, touch wobble works, get_head_angles MCP tool works).
5. **Both `CONFIG_STACKCHAN_SERVO_FEETECH=y` (MIT default) and `CONFIG_STACKCHAN_SERVO_SCSCL=y` (GPL opt-in) build paths** behave identically (the change is in `stackchan.cc` only, not in either driver).

## Refs

- #138 / PR #139 (origin of the safe-fallback seed and Phase 1/2 structure)
- #121 Problem 2 (boot-init speed cap, `BOOT_INIT_MOVE_MS=4000 → 3000` floor + distance-aware re-cast at 15 °/s)
- #137 (snap-suppress hold)
- #100 (SCS0009 bus hang on large-angle pitch reversals — separately surfaced during real-device verification of the initial sweep-based design; see "Implementation history" below)

---

## Implementation history

The initial draft of this Issue proposed a four-corner backlash sweep (Phase 3a/3b/3c/3d) plus a final ReadPos to combine the posture-sync requirement with industry-standard "power-on origin recovery" gear-backlash compensation. Eight rounds of pre-PR codex review hardened the sweep against ServoTask write-back races, stale `WritePos` after axes-idle, per-phase corner-completion guarantees, and a unified `15 °/s` no-stutter angular-speed cap.

Real-device verification of that design hung the SCS0009 bus mid-sweep during Phase 3b (yaw +45° → −45°, pitch 80° → 10°, 6000 ms at 15 °/s). The failure starts ~1 s into the sweep at mid-range coordinates (yaw≈28°, pitch≈68°) — neither at a reversal point nor near an end-stop — and matches the cumulative-overwrite signature of the underlying `ServoTask` design rather than any per-call delta or stall-current effect.

Tracing the boot log against the SCS0009 protocol (`SCSCL_GOAL_TIME` register at SRAM addr 44/45) and three reference implementations (M5Stack official `firmware/main/hal/hal_servo.cpp:80`, `stack-chan/stackchan-arduino` `Stackchan_servo.cpp`, `mongonta0716/Stack-chan-tester-SCS0009`) showed that all three reference designs delegate motion timing to the servo via a single `WritePos(id, pos, time=duration_ms, 0)` per move and poll `ReadMove` for completion. stackchan-mcp's `ServoTask`, by contrast, issues `WritePos(id, pos, 30, 0)` every `MOTION_TICK_MS=20 ms`, which continuously overwrites the SCS0009's in-progress internal motion. The 300+ overwrite frames Phase 3b would have issued is the first place in stackchan-mcp's history where the cumulative effect becomes visible.

Two follow-up actions split out of this discovery:

1. **This Issue / PR** narrows to the cheap-fix scope: drop Phase 3 sweep entirely, keep Phase 0' alone for posture sync. The four-corner sweep was a "nice to have" for backlash compensation, but with the bus-hang trigger it became a net negative.
2. **A separate follow-up Issue (to be filed)** will propose migrating stackchan-mcp's `ServoTask` from the 50 Hz overwrite loop to the single-shot `WritePos(time=duration_ms)` + `ReadMove` polling pattern used by every reference implementation. That migration is the proper root-cause fix for the SCS0009 bus-hang signature originally raised in #1 and tracked through #100; the current Issue's scope is intentionally limited to the posture-sync mismatch.

_Last updated: 2026-05-17_

kisaragi-mochi:
## Summary

Adds a post-Phase 0 re-sync step ("Phase 0'") to `InitializeServo()` that takes a retried `ReadPos` after the boot-init climb and unconditionally overwrites `pitch_motion_.current_deg` / `yaw_motion_.current_deg` with the actual physical position. This eliminates the firmware-side / SCS0009-actual mismatch that the #138 safe-fallback intentionally leaves behind on the PMIC long-press OFF / ON boot path (where Phase 0's `WriteHeadAngles(0, 45, …)` is a no-op of effect because `current_deg` was seeded equal to the target).

Phase 0 is now also distance-aware: its duration is computed from the actual `current_deg → BOOT_INIT_*_DEG` delta at the new `BOOT_INIT_TARGET_DEG_PER_SEC = 15 °/s` cap, floored at `BOOT_INIT_MOVE_MS = 3000 ms` to keep the SCS0009 wake-up latency window covered.

## Why

PR #139 / Issue #138 introduced a safe-fallback that seeds `pitch_motion_.current_deg = BOOT_INIT_PITCH_DEG (=45)` when all pre-init `ReadPos` retries fail (PMIC long-press OFF / ON path). This prevents the end-stop walk during the boot-init interpolation, but leaves the firmware-side `current_deg` several degrees away from the SCS0009's actual physical position. The mismatch magnitude depends on the prior commanded set-point the servo retained across the power cycle — not a fixed value.

Symptom: `move_head(0, 45)` immediately after boot returns `pitch_motion_started: 0` (firmware sees `current_deg == target`, no interpolation starts) and the head does not move, while `get_head_angles` returns the `ReadPos`-based actual position (e.g. 38°). LLM-side clients observe "I asked for 45° but the head didn't move; `get_head_angles` still says 38°" — a contradictory observation hard to attribute correctly without firmware-internals knowledge.

Phase 0' fixes this by re-reading `ReadPos` after the boot-init move has completed (by then the SCS0009 wake-up latency window is well past, so the read reliably succeeds) and re-syncing `current_deg` to the actual physical position. Subsequent `move_head` calls then interpolate from the real starting point.

## Design history

The initial Issue #141 draft proposed a four-corner backlash sweep (Phase 3a/3b/3c/3d) plus a final ReadPos to combine posture sync with industry-standard "power-on origin recovery" gear-backlash compensation. Eight rounds of pre-PR codex review hardened the sweep against ServoTask write-back races, stale `WritePos` after axes-idle, per-phase corner-completion guarantees, and a unified `15 °/s` no-stutter cap. Real-device verification afterwards reproducibly hung the SCS0009 bus mid-sweep during Phase 3b (yaw +45° → −45°, pitch 80° → 10°, 6000 ms at 15 °/s). The failure starts ~1 s into the sweep at mid-range coordinates (yaw≈28°, pitch≈68°) — neither at a reversal point nor near an end-stop — and matches the cumulative-overwrite signature of the underlying ServoTask design rather than any per-call delta or stall-current effect.

Tracing the boot log against the SCS0009 protocol (`SCSCL_GOAL_TIME` register at SRAM addr 44/45) and three reference implementations (M5Stack official `hal_servo.cpp:80`, `stack-chan/stackchan-arduino` `Stackchan_servo.cpp`, `mongonta0716/Stack-chan-tester-SCS0009`) showed that all three reference designs delegate motion timing to the servo via a single `WritePos(id, pos, time=duration_ms, 0)` per move, polling `ReadMove` for completion. stackchan-mcp's `ServoTask` instead issues `WritePos(id, pos, 30, 0)` every `MOTION_TICK_MS=20 ms`, which continuously overwrites the SCS0009's in-progress internal motion. The 300+ overwrite frames Phase 3b would have issued is the first place in stackchan-mcp's history where the cumulative effect becomes visible.

Two follow-up actions:

1. **This PR** (cheap fix): drop Phase 3 entirely, keep Phase 0' alone for posture sync. The four-corner sweep was a "nice to have" for backlash compensation; with the bus-hang trigger it's a net negative.
2. **A separate follow-up Issue** (not yet filed): migrate stackchan-mcp's `ServoTask` from the 50 Hz overwrite loop to the single-shot `WritePos(time=duration_ms)` + `ReadMove` polling pattern used by every reference implementation. That migration root-causes the bus-hang signature originally raised in #1 and tracked through #100; this PR is purely the posture-sync mitigation that #141 was about.

### Build

- [x] `docker run --rm --cpus=4 -v "$PWD":/project -w /project espressif/idf:v5.5.2 python ./scripts/release.py stackchan` — clean rebuild, `xiaozhi.bin` produced

### Real-device verification (`CONFIG_STACKCHAN_SERVO_FEETECH=y`)

- [ ] ESP32-only reset path (`esptool ... write_flash 0x20000 xiaozhi.bin`): Phase 0 / Phase 0' boot log lines emitted in order; final `get_head_angles` returns the actual physical position; subsequent `move_head` calls interpolate from the real `current_deg`
- [ ] PMIC long-press OFF / ON path (hands-off): Phase 0 stays no-op of effect thanks to the #138 safe-fallback seed; Phase 0' re-syncs `current_deg` to the actual position; `move_head(0, 45)` after boot interpolates the head to 45° instead of immediately reporting `pitch_motion_started: 0`
- [ ] No regression on existing motion paths (multiple `move → settle → read` cycles work as before, touch wobble works)

### Pre-PR codex review

`review --base main`, standard. Looped to clean across two design iterations:

- **Phase 3 sweep design (8 rounds)**: ServoTask write-back race, sweep speed cap, stale WritePos race, corner-completion guarantees, blind-sweep on ReadPos failure, per-axis sync skip, Phase 0 distance-aware, post-init ReadPos retry — all reviewed clean by round 8.
- **Phase 3 sweep removed (1 round)**: design simplified to Phase 0 + Phase 0' after real-device sweep verification triggered the SCS0009 bus hang documented above. Round 9: clean — no actionable correctness issues.

## License

MIT FeetechScs canonical build path only (`CONFIG_STACKCHAN_SERVO_FEETECH=y`). No changes to `firmware/main/boards/stackchan/SC*.{cc,h}` (GPL-3.0). All existing license headers preserved.

## Breaking changes

None. New `BOOT_INIT_TARGET_DEG_PER_SEC` constant is additive. Existing `BOOT_INIT_YAW_DEG=0` / `BOOT_INIT_PITCH_DEG=45` / `BOOT_INIT_MOVE_MS=3000` are kept; Phase 0 now uses `BOOT_INIT_MOVE_MS` as a floor for the distance-computed duration rather than as the absolute value. `WriteHeadAngles` / `set_head_angles` / `move_head` / `get_head_angles` and the `servo_motion` task all behave identically for normal user-driven motion.

## Boot-time budget

Worst-case Phase 0 climb (45° travel at 15 °/s) is 3.0 s; post-Phase 0 ReadPos retry plus diagnostic log adds <0.5 s. Total boot-init duration ~3.5 s before MCP tools become responsive — significantly faster than any of the Phase 3 sweep variants considered earlier in the review history.

## Caveats / Future work

- The SCS0009 bus-hang signature reproduced during the Phase 3 sweep is the visible head of a deeper design issue in `ServoTask` (50 Hz `WritePos` overwrite of the SCS0009's in-progress internal motion). A separate follow-up Issue will propose migrating to the reference single-shot + `ReadMove` polling pattern. This PR is intentionally scoped to the Issue #141 posture-sync requirement only and leaves the `ServoTask` overwrite loop untouched.
- This PR does not change the semantic of `BOOT_INIT_PITCH_DEG=45`. The `listen()` look-up pose introduced by #96 happens to use the same angle; whether to split them into separately-named constants is a future cleanup that can ride on top of the `ServoTask` migration.

## Related issues

Closes #141.
Refs #138 / PR #139, #137, #117, #121 Problem 1, #96, #140, #100.

### DIFF
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -707,15 +707,24 @@ private:
     // positioning timing established in mongonta0716/stackchan-arduino
     // attachServos().
     //
-    // Issue #121 Problem 2: BOOT_INIT_MOVE_MS=4000 (was 1000) drops the
-    // effective angular speed for the post-power-on climb from ~45°/s
-    // to ~11°/s. On-device feedback identified the original 1-second
-    // duration as startling ("ブルンっ" / audible servo stress) on the
-    // CoreS3 + SCS0009 hardware. The 100 ms post-settle vTaskDelay in
-    // InitializeServo() is unchanged. The separate "unintended downward
-    // drop on power-on" (#121 Problem 1) is addressed by the
-    // snap-suppress hold in InitializeServo() Phase 1a (PR #137) and
-    // the ReadPos retry + safe-fallback seed in this file (#138).
+    // Speed policy history (#121 Problem 2 -> #141 follow-ups):
+    // - #121 Problem 2 originally raised BOOT_INIT_MOVE_MS from 1000 ms
+    //   (the historical default that produced a startling "ブルンっ" boot
+    //   motion) to 4000 ms (~11°/s on a 45° climb).
+    // - Real-device verification then showed even 11°/s reads as
+    //   perceptibly fast for the first boot-time servo motion, so the
+    //   Phase 0 climb is pinned to an angular-speed cap of
+    //   BOOT_INIT_TARGET_DEG_PER_SEC=15 deg/s. The WriteHeadAngles call
+    //   sizes its duration from the actual yaw / pitch deltas so this
+    //   cap holds on every axis. On the PMIC OFF/ON path Phase 0 stays
+    //   a no-op of effect (the #138 safe-fallback seed makes start_deg
+    //   == target_deg), so the BOOT_INIT_MOVE_MS budget simply elapses
+    //   without WritePos movement.
+    // The 100 ms post-settle vTaskDelay in InitializeServo() is
+    // unchanged. The separate "unintended downward drop on power-on"
+    // (#121 Problem 1) is addressed by the snap-suppress hold in
+    // InitializeServo() Phase 1a (PR #137) and the ReadPos retry +
+    // safe-fallback seed in this file (#138).
     //
     // Issue #138: promoted from local block scope inside
     // InitializeServo() to class-level static constexpr so the
@@ -730,7 +739,37 @@ private:
     // mid-sequence.
     static constexpr int BOOT_INIT_YAW_DEG = 0;
     static constexpr int BOOT_INIT_PITCH_DEG = 45;
-    static constexpr uint32_t BOOT_INIT_MOVE_MS = 4000;
+    // BOOT_INIT_MOVE_MS=3000: minimum duration of the Phase 0 climb.
+    // Used as a floor so the boot-init `WriteHeadAngles(0, 45, X)`
+    // always elapses at least this long — required on the PMIC OFF/ON
+    // path where the #138 safe-fallback seed makes Phase 0 a no-op of
+    // effect, and the BOOT_INIT_MOVE_MS budget instead serves to span
+    // the SCS0009 wake-up latency window so the post-init ReadPos
+    // (Phase 0') lands well past it. The actual Phase 0 duration is
+    // computed at call time from the current_deg → BOOT_INIT_*
+    // deltas at BOOT_INIT_TARGET_DEG_PER_SEC=15 deg/s, then floored at this
+    // constant; e.g. on the ESP32-only reset path with a yaw-90° prior
+    // set-point, Phase 0 needs 6000 ms to honour the speed cap while a
+    // yaw-0 prior is rounded up to this 3000 ms floor. This is the
+    // no-stutter Smooth lower bound established under #121 Problem 2
+    // (Issue #121 / PR #125 history: 1000 -> 4000 was a partial step
+    // toward this; on-device feedback under #141 verification confirmed
+    // 15 deg/s is the speed at which the ServoTask MOTION_TICK_MS=20 ms
+    // interpolation stops being perceptible as individual position
+    // jumps without sliding into a startling regime). Boot-time budget
+    // is intentionally not optimised: operator safety and avoiding
+    // mechanical stress take precedence over shaving seconds off the
+    // initialization duration.
+    //
+    // On the PMIC OFF/ON path Phase 0 stays a no-op of effect
+    // because the #138 safe-fallback seeds current_deg to
+    // BOOT_INIT_PITCH_DEG, making start_deg == target_deg; the
+    // BOOT_INIT_MOVE_MS budget then simply elapses without WritePos
+    // movement.
+    static constexpr uint32_t BOOT_INIT_MOVE_MS = 3000;
+    // Single-source Phase 0 speed cap. 15 deg/s is the no-stutter Smooth
+    // lower bound (#121 Problem 2 + #141 verification).
+    static constexpr int BOOT_INIT_TARGET_DEG_PER_SEC = 15;
 
     static int YawDegToPos(int deg) {
         int pos = 460 + deg * 16 / 5;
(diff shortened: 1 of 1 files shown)


kisaragi-mochi:
Filed the follow-up motion-subsystem refactor as #143 — that one root-causes the 50Hz `WritePos` overwrite mechanism (the same mechanism that surfaced as the Phase 3b bus hang during this PR's real-device verification). This PR stays narrowly scoped to the Issue #141 posture-sync requirement; #143 will retire the underlying 50Hz interpolation loop in a separate, opt-in-gated migration.

Issue #161, opened 2026-05-17
bug(firmware): HostInterpolationMotionDriver::Tick releases motion_mutex_ between snapshot and WritePos — pre-bus stale-command race


kisaragi-mochi:
## Summary

`HostInterpolationMotionDriver::Tick()` releases `motion_mutex_` between
its `AxisMotion` snapshot and the `scs_bus_mutex_`-protected `WritePos`
calls. If `StartMove` (same-ms or otherwise) or `InitializeServo`'s
Phase 0' direct re-sync lands during that release window, the next
`WritePos` is issued from a stale snapshot — the physical servo
receives the old motion's intermediate position even though the
freshly-written `AxisMotion` (and now the `request_token`) tell
`Tick()` to skip the post-bus commit.

In other words, the post-bus `request_token`-equality guard (Issue
#158 / PR for #158) prevents the **state divergence** (`current_deg` /
`moving` corruption) but does not prevent the **physical bus
contamination** — the stale `WritePos` has already been sent by the
time the freshness guard kicks in.

## Race window (theoretical)

1. `Tick()` takes `motion_mutex_`, copies `yaw_motion_` / `pitch_motion_`
   into `yaw_local` / `pitch_local`, releases `motion_mutex_`.
2. `Tick()` computes `new_*_current` from `yaw_local` / `pitch_local`
   (linear or spring), takes `scs_bus_mutex_`, and starts the
   `WritePos` sequence.
3. `StartMove` (or Phase 0' direct re-sync) takes `motion_mutex_`,
   updates `AxisMotion` (including `request_token` via the new path
   from Issue #158), releases `motion_mutex_`.
4. `Tick()`'s `WritePos` for the OLD interpolation lands on the
   SCS0009 bus before the new dispatcher has a chance to issue its
   first `WritePos`. The physical head briefly tracks the cancelled
   motion's intermediate position before the new motion takes over.

Impact:
- Brief physical excursion past the new target during the cancellation.
- During Phase 0' boot re-sync: the `ReadPos` sample is captured, but
  one more stale `WritePos` may follow, contradicting Phase 0''s
  documented intent of "no additional boot-time WritePos commands on
  a degraded bus".

Severity is **medium**: rare in steady-state operation (depends on
`Tick()` being mid-cycle when `StartMove` / Phase 0' fires), and
self-correcting within one `MOTION_TICK_MS` (20 ms) cycle. Not a hang
[...]

Issue #160, opened 2026-05-17
bug(firmware): Phase 0' re-sync does not bump request_token — same-token race vs both motion drivers


kisaragi-mochi:
## Summary

`InitializeServo()`'s Phase 0' post-init `ReadPos` re-sync (lines
2509-2604 of `firmware/main/boards/stackchan/stackchan.cc`) directly
mutates `yaw_motion_` / `pitch_motion_` fields (`current_deg`,
`target_deg`, `start_deg`, `moving`, `move_start_ms`) under
`motion_mutex_` without going through either motion driver's
`StartMove` path. Neither `HostInterpolationMotionDriver::next_request_token_`
nor `ServoDelegatedMotionDriver::next_request_token_` is bumped, so
`AxisMotion::request_token` stays unchanged across the re-sync.

After Issue #158 / PR #N (port `request_token` guard to host
interpolation), `HostInterpolationMotionDriver::Tick()`'s post-bus
write-back uses `request_token` equality as its freshness guard. If a
`Tick()` iteration has already snapshotted `AxisMotion` (capturing the
pre-resync `request_token`) and a Phase 0' re-sync lands while the bus
I/O is in flight (between `motion_mutex_` release and reacquire), the
post-bus guard sees the same `request_token` on both sides and commits
the stale `current_deg` / `moving=false` over the just-re-synced
ReadPos-based state.

The previous `move_start_ms`-based guard happened to reject this race
because Phase 0' overwrites `move_start_ms` to `boot_init_end_tick *
portTICK_PERIOD_MS` (a different value from the staging timestamp).
Swapping in `request_token` removes that incidental protection.

## Why this is filed as a separate Issue

Issue #158 scope is "port the `request_token` guard from the delegated
driver to the host-interpolation driver" — a single-file, ~15-line
change inside `HostInterpolationMotionDriver`. Plumbing `request_token`
invalidation from `Board::InitializeServo` (which is the owner of the
Phase 0' re-sync logic) into both motion drivers requires either:

- adding a new public method on the `MotionDriver` interface
  (e.g. `InvalidateAxisTokens()`), implemented by both
  `HostInterpolationMotionDriver` and `ServoDelegatedMotionDriver`, or
- moving the `next_request_token_` counter ownership out of the
  motion drivers and into the board, or
- factoring Phase 0' itself into a motion-driver-aware "external state
  override" path with explicit token handshake.

Each of these is a design call that warrants its own review and its
[...]

Issue #158, opened 2026-05-17
bug(firmware): HostInterpolationMotionDriver write-back race — ms-resolution move_start_ms guard collides on same-ms StartMove


kisaragi-mochi:
## Summary

`HostInterpolationMotionDriver::Tick()` writes back per-axis `current_deg` / `moving` state using only `move_start_ms` as its freshness guard. `StartMove` stamps that field from `esp_timer_get_time() / 1000`, so two `StartMove` calls landing in the same millisecond share the guard value — a stale `Tick` snapshot taken before the newer command can then pass the guard after bus I/O and overwrite the newer move's state.

`ServoDelegatedMotionDriver` already protects against the same class of race using a per-driver monotonic `request_token` counter (introduced as part of PR #146 / #154 Phase 2). `HostInterpolationMotionDriver` should adopt the same pattern (or move to µs-resolution timestamps) for parity.

## Background

Raised by `codex adversarial-review` against `kisaragi-mochi/stackchan-mcp` Issue #152 Phase 3 (commit `89bbdf6` on branch `issue-152-phase-3-spring-animation`). The replacement of linear interpolation with `smooth_ui_toolkit` spring physics did not introduce this bug — the ms-resolution freshness guard predates Phase 3 — but Phase 3 also did not address it, and now both motion drivers should ideally share the same race protection.

## Race window (theoretical)

1. Caller A invokes `StartMove(yaw_A, pitch_A, dur_A)`. `move_start_ms = T_ms`.
2. `Tick` enters and snapshots `yaw_motion_` / `pitch_motion_` into local `yaw_local` / `pitch_local`.
3. Before `Tick` writes back, Caller B invokes `StartMove(yaw_B, pitch_B, dur_B)` within the same millisecond. `move_start_ms = T_ms` (unchanged because of millisecond rounding).
4. `Tick` proceeds with bus I/O for A's interpolation, then takes `motion_mutex_` to write back. The freshness guard `yaw_motion_.move_start_ms == yaw_local.move_start_ms` evaluates `true` (same `T_ms`), so A's stale `current_deg` / `moving` overwrites B's freshly written state.

Impact:
- Reported `current_deg` and `moving` flag diverge briefly from the actual commanded move.
- For the spring path, `axis_anim_` state (set in B's `StartMove`) and the `AxisMotion` state can be observed inconsistently across `motion_mutex_` boundaries.

[...]

Issue #152, opened 2026-05-17
feat(firmware): complete 'hardware-from-StackChan-official' architecture — import m5stack/StackChan motion subsystem (smooth_ui_toolkit spring physics + per-servo independent update)


kisaragi-mochi:
## Context — completing an unfinished architecture choice

stackchan-mcp's firmware is the result of an interim approach:

- **Hardware spec** (CoreS3 + SCS0009 ×2, GPIO mapping, etc.) was sourced from the m5stack/StackChan official firmware (e.g. `firmware/main/hal/hal_servo.cpp` was the original source for the SCS0009 pin/baud information).
- **Software for motion control was hand-rolled on top of the xiaozhi-esp32 fork**, NOT imported from m5stack/StackChan.

The hand-rolled path was a tactical decision when MCP integration and voice AI were the priority. It is now reaching its limits:

- Linear interpolation in `HostInterpolationMotionDriver` was the documented contributor to the SCS0009 bus-hang surface tracked under #100 / #137 / #138 / #139 / #141.
- PR #146's delegated path mitigates substantially but leaves a residual hang trigger when two-axis simultaneous dispatch combines with pitch end-stop proximity or cumulative session load (E1–E6 verification in PR #146 body).
- **m5stack/StackChan ships the same hardware widely with no equivalent bus-hang report.** Reading their motion subsystem makes the reason visible:
  - `smooth_ui_toolkit::SpringAnimation` for frame-by-frame natural motion (small per-`WritePos` deltas; reference: `firmware/main/stackchan/motion/servo.cpp:13-67` in m5stack/StackChan).
  - **Per-Servo independent `update()` tick** (no two-axis simultaneous-dispatch race; `Motion::move()` and `Motion::moveWithSpeed()` simply forward to each `Servo`, which runs its own 50 Hz update loop).

This Epic tracks completing the architectural migration: replace the hand-rolled motion subsystem with the m5stack/StackChan motion architecture, so the "hardware from StackChan official" choice extends to the software layer that owns the same hardware.

## License (verified)

- **m5stack/StackChan motion subsystem**: MIT (SPDX-FileCopyrightText: 2026 M5Stack Technology CO LTD on the relevant files).
- **smooth_ui_toolkit** (`Forairaaaaa/smooth_ui_toolkit`): MIT (Copyright (c) 2023 Forairaaaaa, 153 stars, 2026-05-13 updated).

[...]

PR #164, opened 2026-05-17
feat(firmware/gateway): #163 add set_servo_torque MCP tool for Phase 4 idle-release probe

### LINKED ISSUE #163: feat(firmware): MCP tool 'set_servo_torque' — observability probe for #152 Phase 4 idle-release decision
## Summary

Add a runtime MCP tool `self.robot.set_servo_torque(yaw_enabled, pitch_enabled)` so the SCS0009 torque on each axis can be toggled at runtime. The immediate purpose is to physically observe head behavior with torque OFF on real hardware — does the head hold in place via static friction, or does it drift toward an end-stop under gravity? — before designing Issue #152 Phase 4 (auto-torque-release on idle).

Phase 4's "release torque after N seconds of idle" hinges on this observation. Real-device evidence so far (M5Stack stack-chan + SCS-0009) suggests static friction holds the head at the last commanded position even during PMIC-OFF torque drop (head remains at ~45° rather than gravity-dropping to 0°), but that observation has only been incidental. A runtime probe lets the maintainer toggle torque deliberately and watch the result without flashing a special test firmware.

## Background

`MotionDriver` already calls `EnableTorque(id, 1)` during `InitializeServo` (and the SCS0009 firmware path is bus-frame compatible across the FeetechScs MIT driver and the legacy GPL SCSCL driver). Exposing `EnableTorque` as an MCP tool is a tiny additive change on the firmware side and is fully covered by the gateway's dynamic MCP-tool schema fetch (no gateway-side code change required, per the existing `set_head_angles` / `get_head_angles` precedent).

## Proposed change

Single-file change in `firmware/main/boards/stackchan/stackchan.cc`:

- Add an `mcp_server.AddTool("self.robot.set_servo_torque", ...)` registration alongside the existing `set_head_angles` / `get_head_angles` / `set_avatar` ... tools. Tool naming follows the same `self.robot.<verb>` convention.
- Properties:
  - `yaw_enabled` (bool): true → `EnableTorque(SERVO_YAW_ID, 1)`, false → `EnableTorque(SERVO_YAW_ID, 0)`.
  - `pitch_enabled` (bool): same for pitch.
- Handler:
  - Take `scs_bus_mutex_` (consistent with other bus-touching handlers).
  - Call `ScsBus::EnableTorque` for each axis.
  - Release `scs_bus_mutex_`.
  - Log the action (`ESP_LOGI`) and the per-axis bus return code.
  - Return a small JSON with the bus return codes plus the requested states.
- No interaction with `MotionDriver` state (`AxisMotion::request_token`, `move_start_ms`, etc.). The probe deliberately sidesteps the motion abstraction — it is a low-level bus operation, not a motion command. Re-enabling torque does NOT trigger a move; the next `set_head_angles` or wobble does.
- No change to default startup behavior. After boot, torque is enabled by `InitializeServo` as today; this tool only takes effect when explicitly invoked.

## Acceptance

- [ ] Firmware build (`python ./scripts/release.py stackchan`) succeeds and the new tool appears in the gateway's dynamic tool listing after MCP reconnect.
- [ ] `set_servo_torque(yaw_enabled=false, pitch_enabled=false)` issues two `EnableTorque(..., 0)` frames and returns success.
- [ ] After the OFF call, `get_head_angles` still returns sensible degrees (SCS0009 `ReadPos` is independent of torque state on this servo).
- [ ] After the OFF call, the physical head behavior is observable: maintainer notes whether the head holds via friction or drifts under gravity.
- [ ] `set_servo_torque(yaw_enabled=true, pitch_enabled=true)` re-enables torque and `set_head_angles(0, 45)` resumes normal interpolation from the post-OFF physical position.
- [ ] License boundary unchanged: only `stackchan.cc` (MIT) is modified; `SC*.{cc,h}`, `INST.h`, `SCServo.h` (GPL-3.0) untouched. `ScsBus::EnableTorque` is used as the abstraction, which lives on the MIT FeetechScs canonical path.

## Out of scope (tracked separately)

- The actual idle-detection + auto-release timer logic — Issue #152 Phase 4.
- A "safe" wrapper that suppresses `set_head_angles` while torque is OFF — Phase 4 design will decide whether to gate or to auto-re-enable.
- Persistence of the torque state across reboots — Phase 4 will decide whether the post-release state survives a PMIC cycle.

## Notes

- The `set_servo_torque` tool may be retained long-term as a diagnostic / power-management primitive, or folded into Phase 4's auto-release plumbing. The Phase 4 PR can decide.
- Surfaces under MCP, so the maintainer can drive it from Claude Code / any MCP-aware client without a physical button or hidden serial command.

## Related

- Issue #152 — origin: complete the m5stack/StackChan motion subsystem migration. Phase 4 is the optional final piece.
- Issue #100 — SCS0009 bus hang root tracker. Torque OFF removes torque current and is NOT expected to trigger the #100 hang surface, but the observation is part of the spike value.
- PR #146 / PR #154 — origin of the per-axis driver abstraction; this probe deliberately sidesteps the abstraction since torque toggle is a lower-level bus operation.

kisaragi-mochi:
## Summary

Add a new MCP tool `self.robot.set_servo_torque(yaw_enabled,
pitch_enabled)` to firmware + gateway. The tool toggles SCS0009 torque
on each axis independently via `ScsBus::EnableTorque(id, enable)`. It
is a **diagnostic probe** for #152 Phase 4 (auto-torque-release on
idle): the maintainer can call it to physically observe whether the
head holds via static friction or drifts under gravity with torque off
on real hardware, before the auto-release timer is designed.

### What the firmware does on disable

When at least one axis is disabled, the handler runs (under the same
`motion_mutex_`-protected critical section, **before** the
`EnableTorque` bus frame):

1. `axis_motion_.moving = false` for the disabled axis.
2. `axis_motion_.position_unknown = true`.
3. `motion_driver_->InvalidateAxisToken(axis_id)` — bumps the freshness
   token (Issue #158 / PR #162 mechanism) so any in-flight Tick
   snapshot is invalidated.
4. `servo_wobble_active_.store(false)` and `servo_wobble_step_.store(0)`
   — cancels any in-flight wobble sequence (mirrors
   `WriteHeadAngles`'s wobble-cancel block).

Doing the state cancellation **before** the `EnableTorque` bus frame
narrows the window in which a stale Tick can emit one last `WritePos`
after `scs_bus_mutex_` is released.

### What the firmware does on enable

Nothing beyond the bus frame itself. The intended workflow is: the
caller reuses `get_head_angles` (which reads the SCS0009 raw position,
independent of torque state) plus a fresh `set_head_angles` to
re-anchor the firmware's internal `current_deg` if needed. `EnableTorque`
re-enable does not trigger any motion.

### Gateway exposure

`gateway/stackchan_mcp/stdio_server.py` adds the matching `Tool(...)`
entry in `list_tools()` (boolean inputSchema for each axis) and the
matching `tool_map` entry (`"set_servo_torque" → "self.robot.set_servo_torque"`).
The `mcp_router.py` subset path is unchanged; this probe is exposed on
the standard stdio MCP path.

### Code-level

- [x] firmware: `python ./scripts/release.py stackchan` (Docker
      `espressif/idf:v5.5.2`, canonical
      `CONFIG_STACKCHAN_SERVO_FEETECH=y` build). Builds clean; new tool
      symbol present in the linked binary.
- [ ] gateway: no automated test added — the new entry follows the same
      shape as `set_blink` (which has no dedicated test) and is dispatched
      through the same `tool_map` mechanism.

### Hardware

- [ ] Device boots without crash, existing MCP tools unaffected —
      handed to maintainer
- [ ] **Idle-state probe** (primary use case): `set_servo_torque(false,
      false)` from rest; observe whether the head holds (static
      friction) or drifts (gravity). `set_servo_torque(true, true)` to
      re-enable; verify subsequent `set_head_angles` resumes normal
      motion from the post-OFF physical position via the
      `get_head_angles` + `set_head_angles` re-anchor pattern — handed
      to maintainer
- [ ] **Mid-motion disable**: call `set_servo_torque(false, false)`
      during an active `set_head_angles` interpolation; verify the
      ServoTask stops emitting `WritePos` ticks (any leak is the
      carved-out `#161` pre-bus race) — handed to maintainer
- [ ] **Mid-wobble disable**: trigger a touch wobble, then disable
      torque mid-wobble; verify the wobble sequence does not resume
      after `set_servo_torque(true, true)` — handed to maintainer

## Pre-PR codex review

`review --base main`, four rounds. Findings addressed inline this PR:

- R1 P2: cancel `MotionDriver` state on disable → addressed (motion
  state cancellation block).
- R2 P2: expose the tool on the gateway side → addressed
  (`stdio_server.py` additions).
- R2 P2: stale `WritePos` ordering — disable bus frame fired before
  motion state cancel → addressed (reordered: cancel first, then bus
  frame).
- R2 P2: delegated path `pending_dispatch_` not cleared → carved out
  under **#160**, documented inline in the handler.
- R3 P2: in-flight wobble cancel → addressed (`servo_wobble_active_`
  cleared in the same critical section).
- R4 P2: pre-bus stale `WritePos` window in `Tick()` (the
  cancellation-before-bus-frame ordering narrows it but does not
  eliminate it on the host-interpolation path) → same pre-bus race
  surface as **#161** (pre-existing). Carved out; this probe does not
  introduce or widen it.
- R4 P2: `HostInterpolationMotionDriver::StartMove` does not consult
  `position_unknown`, so a same-target `set_head_angles` after a
  torque-off + hand-push can be no-op-optimized → carved out under
  **#160** (the broader `position_unknown` honoring across both drivers
  is part of the canonicalization).

## Known limitations

- **Pre-bus stale `WritePos` race** in
  `HostInterpolationMotionDriver::Tick()` when torque is disabled
  mid-interpolation: same race surface as
  [#161](https://github.com/kisaragi-mochi/stackchan-mcp/issues/161),
  pre-existing, not introduced by this PR. The cancellation-before-bus-
  frame ordering in this handler narrows the window; closing it
  requires the same pre-bus token gate tracked under #161.
- **Same-target re-dispatch after a torque-off hand-push**: the
  default host-interpolation path does not honor `position_unknown` in
  `StartMove`, so the documented "re-anchor with
  `get_head_angles` + `set_head_angles`" workflow is required after a
  manual reposition under torque-off. Honoring `position_unknown` in
  `HostInterpolationMotionDriver::StartMove` is part of the broader
  canonicalization tracked under
  [#160](https://github.com/kisaragi-mochi/stackchan-mcp/issues/160).
- **Delegated path `pending_dispatch_`** is not cleared by the
  cancellation block; the `AxisServo` private state machine refactor
  is tracked under #160.

## License

`firmware/main/boards/stackchan/SC*.{cc,h}`, `INST.h`, `SCServo.h`
(GPL-3.0) untouched. Firmware change is confined to MIT-licensed
`stackchan.cc`; `ScsBus::EnableTorque` is the MIT-canonical
FeetechScs-path abstraction. Gateway change is in
`gateway/stackchan_mcp/stdio_server.py` (MIT).

## Breaking changes

None. The new MCP tool is additive on both firmware and gateway sides;
existing tools and startup behavior are unchanged.

## Related issues

Closes #163
Refs #152 (Phase 4 auto-release follow-up that this probe informs)
Refs #160 (delegated-path / host-interpolation cancellation
canonicalization — covers F4/F6 carve-outs above)
Refs #161 (pre-bus stale-`WritePos` race — covers F5 carve-out above)

### DIFF
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -3945,6 +3945,120 @@ private:
                 return result;
             });
 
+        mcp_server.AddTool(
+            "self.robot.set_servo_torque",
+            "Enable or disable SCS0009 servo torque on the yaw / pitch axes "
+            "independently. Disabling torque stops motor current on that axis; "
+            "the head holds via static friction (no motion is commanded). "
+            "On disable, the corresponding axis's MotionDriver state is reset "
+            "(moving=false, position_unknown=true, request token invalidated) "
+            "so a stale interpolation cannot resume on the bus and a "
+            "subsequent same-target set_head_angles is re-dispatched rather "
+            "than no-op-optimized. Re-enabling torque does NOT trigger a "
+            "move -- the next set_head_angles or wobble call will. Returns "
+            "the per-axis bus return codes. Diagnostic / power-management "
+            "primitive; auto release on idle is tracked separately under "
+            "#152 Phase 4.",
+            PropertyList({Property("yaw_enabled", kPropertyTypeBoolean),
+                          Property("pitch_enabled", kPropertyTypeBoolean)}),
+            [this](const PropertyList& properties) -> ReturnValue {
+                bool yaw_enabled = properties["yaw_enabled"].value<bool>();
+                bool pitch_enabled = properties["pitch_enabled"].value<bool>();
+
+                // Cancel MotionDriver state for any axis whose torque is
+                // being disabled BEFORE issuing the EnableTorque bus
+                // frame. Without this ordering, ServoTask could snapshot
+                // motion state while moving=true, wait on scs_bus_mutex_
+                // through our EnableTorque(OFF), then fire one stale
+                // WritePos on the freshly-disabled axis right after we
+                // release scs_bus_mutex_ (silently absorbed while torque
+                // is off, then resuming audibly when torque comes back).
+                // Performing the cancellation first lets the next Tick
+                // observe moving=false / a fresh request_token before it
+                // commits any WritePos. position_unknown forces the
+                // delegated path to re-dispatch the next move
+                // (HostInterpolation ignores the flag, but the
+                // moving=false + token bump already block resumption
+                // there). Re-enable is treated as a user-driven re-
+                // anchor: the caller is expected to use get_head_angles
+                // + set_head_angles to re-establish a verified position
+                // if needed.
+                //
+                // Known carve-out (#160): on the delegated path,
+                // ServoDelegatedMotionDriver does not override
+                // InvalidateAxisToken, and the per-axis AxisServo private
+                // pending_dispatch_ / retry-state is not cleared by this
+                // path. A subsequent Update() tick can therefore still
+                // re-stage a WritePos on the freshly-disabled axis from
+                // pending state captured before this call. Full
+                // delegated-path cancellation requires the same refactor
+                // tracked under #160 and is out of scope for this probe.
+                if (motion_driver_ != nullptr && motion_mutex_ != nullptr &&
+                    (!yaw_enabled || !pitch_enabled)) {
+                    xSemaphoreTake(motion_mutex_, portMAX_DELAY);
+                    if (!yaw_enabled) {
+                        yaw_motion_.moving = false;
+                        yaw_motion_.position_unknown = true;
+                        motion_driver_->InvalidateAxisToken(SERVO_YAW_ID);
+                    }
+                    if (!pitch_enabled) {
+                        pitch_motion_.moving = false;
+                        pitch_motion_.position_unknown = true;
+                        motion_driver_->InvalidateAxisToken(SERVO_PITCH_ID);
+                    }
+                    // Also cancel any in-flight servo wobble sequence.
+                    // Wobble drives both axes through ServoTaskMain's
+                    // ServoWobbleStepAdvance after Tick; without
+                    // clearing the active flag here, the next tick
+                    // would observe moving==false and stage the next
+                    // wobble StartMove with torque off, enqueuing fresh
+                    // WritePos targets that would run on re-enable.
+                    // Mirrors the wobble-cancel sequence in
+                    // WriteHeadAngles.
+                    servo_wobble_active_.store(false);
+                    servo_wobble_step_.store(0);
+                    xSemaphoreGive(motion_mutex_);
+                }
+
+                int r_yaw = -1;
+                int r_pitch = -1;
+                if (servo_ok_ && scs_bus_mutex_ != nullptr) {
+                    xSemaphoreTake(scs_bus_mutex_, portMAX_DELAY);
+                    r_yaw = scs_bus_.EnableTorque(SERVO_YAW_ID,
+                                                  yaw_enabled ? 1 : 0);
+                    r_pitch = scs_bus_.EnableTorque(SERVO_PITCH_ID,
+                                                    pitch_enabled ? 1 : 0);
+                    xSemaphoreGive(scs_bus_mutex_);
+                }
+
+#if CONFIG_STACKCHAN_SERVO_FEETECH
+                bool yaw_ok = (r_yaw == 0);
+                bool pitch_ok = (r_pitch == 0);
+#else
+                bool yaw_ok = (r_yaw > 0);
+                bool pitch_ok = (r_pitch > 0);
+#endif
+
+                cJSON* root = cJSON_CreateObject();
+                cJSON_AddBoolToObject(root, "yaw_enabled", yaw_enabled);
+                cJSON_AddBoolToObject(root, "pitch_enabled", pitch_enabled);
+                cJSON_AddNumberToObject(root, "yaw_bus_return", r_yaw);
+                cJSON_AddNumberToObject(root, "pitch_bus_return", r_pitch);
+                cJSON_AddBoolToObject(root, "servo_ok", servo_ok_);
+                cJSON_AddBoolToObject(root, "ok", servo_ok_ && yaw_ok && pitch_ok);
+                if (!servo_ok_) {
+                    cJSON_AddStringToObject(root, "error",
+                                            "Servo bus not initialized.");
+                }
+                ESP_LOGI(TAG,
+                         "set_servo_torque: servo_ok=%d yaw_enabled=%d (r=%d) "
+                         "pitch_enabled=%d (r=%d)",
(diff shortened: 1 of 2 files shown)


kisaragi-mochi:
## Paused — to be resumed after #160 + Phase 4

During real-device verification (2026-05-18 session), this probe surfaced two integration issues that must be resolved before the tool delivers its intended value:

1. **ServoDelegated path does not persistently disable torque.**
   `set_servo_torque(false, false)` momentarily releases torque (a brief backlash motion was visually observed) but the `AxisServo::Update()` tick immediately re-clamps via its private `pending_dispatch_` state. The board-level `motion_mutex_` block (cancel + `position_unknown=true` + `InvalidateAxisToken`) does not propagate into the per-`AxisServo` state because `InvalidateAxisToken` is a no-op on `ServoDelegatedMotionDriver` by design — exactly the partial-plug problem carved out as #160.
2. **`bus_return=-1` Ack-semantics suspicion.**
   Both enable and disable calls return `yaw_bus_return=-1 / pitch_bus_return=-1` with `ok=false`, despite the disable path producing real bus-level effect (the backlash motion above). The `#if CONFIG_STACKCHAN_SERVO_FEETECH` Ack-handling branch likely needs review for the ServoDelegated wire-level return semantics.

In addition, the session confirmed root cause for the broader "long-uptime hang" pattern as #165 (SCS0009 cumulative-WritePos protection mode). Phase 4 idle auto-release is the primary mitigation candidate, and this probe will be the natural observability tool for that work.

### Resume conditions

This PR will be reopened (or replaced by a fresh PR off `main`) when:

- #160 lands or is otherwise addressed such that `set_servo_torque(false)` persists on the ServoDelegated path, AND
- Phase 4 (Issue #152 outstanding scope) is being designed/implemented and benefits from this probe

### Branch retention

The branch `issue-163-set-servo-torque-probe` will be kept on the remote so this work can be reopened directly. No `--delete-branch` on close.

### Related

- #163 — parent Issue (probe intent)
- #160 — partial-plug blocker
- #165 — root cause confirmed (cumulative-WritePos protection mode)
- #152 — Phase 4 parent

kisaragi-mochi:
Closing for now per the pause note above. Branch `issue-163-set-servo-torque-probe` is retained on the remote for direct resume.

kisaragi-mochi:
Superseded by PR #167 (rebased onto `main` at `bb27431` which includes PR #166; head ref moved post-close blocked reopen, so opened as a fresh PR with the same diff). Both Paused-note items are resolved — see the **Update on resume** section in #167 for details.

PR #186, opened 2026-05-19
fix(firmware/wifi): retry first-attempt association failure with delay

maha0525:
## Symptom

When a user runs through the captive-portal Wi-Fi setup, the first
submission of SSID/Password always fails. Submitting the **same**
credentials a second time (~8s later) succeeds. This makes the setup
UX confusing — users frequently think they typed something wrong.

## Root cause

In APSTA mode the captive-portal session is on the AP beacon channel
(typically 1), and the target home AP is on a different channel
(e.g. 10 in my test environment). When `ConnectToWifi` triggers,
esp-wifi performs a Channel Switch Announcement to move both AP and
STA to the home AP's channel, then immediately issues an association
request.

The home AP responds with "Association Response status=30 (Refused
Temporarily)" plus a Comeback Time (observed ~1.1s on a Buffalo
router) because its own state hasn't settled for the new station.

The ESP-IDF wifi driver's `failure_retry_cnt` issues re-association
attempts back-to-back (within a few ms) and does **not** honor the
802.11 Comeback Time, so every driver-internal retry is refused the
same way and the first `ConnectToWifi()` call returns failure. By the
time the user clicks "submit" a second time (~8s later) the AP has
fully settled and association succeeds on the first try — which is
why users observe "it always fails the first time, then works".

## Fix

Add an upper-level retry loop in `ConnectToWifi` with a 3-second delay
between attempts. When an attempt fails, wait long enough for the
comeback timer + AP state to settle, then retry once. This produces a
single user-visible success path instead of forcing the user to
resubmit.

The driver-internal `failure_retry_cnt` is kept at its default 1; the
retry logic lives in the upper layer where the delay is easy to
express.

## Verification

Tested on M5Stack CoreS3 (stackchan board) against a Buffalo
Buffalo-G-E640 router (WPA2-PSK, channel 10).

**Before** (first attempt fails, user has to resubmit):

    Connecting to WiFi Buffalo-G-E640
    wifi: Association refused temporarily, comeback time 1124 TUs
    wifi: Association refused too many times, max allowed 1
    Failed to connect to WiFi Buffalo-G-E640
    [user has to resubmit, ~8s later succeeds]

**After** (single user-visible success after ~7s):

    Connecting to WiFi Buffalo-G-E640 (attempt 1/2)
    wifi: Association refused temporarily, comeback time 1124 TUs
    Attempt 1/2 failed — will retry
    WiFi attempt 2/2 after 3000 ms delay
    Connecting to WiFi Buffalo-G-E640 (attempt 2/2)
    Connected to WiFi Buffalo-G-E640

## Limitations

- Observed on a Buffalo router only; the underlying 802.11 spec
  allows any compliant router to defer association similarly
  (especially with MFP / 802.11w enabled), so the fix should help
  broadly but has only been verified against this one AP model.
- Environments where the first attempt already succeeds are
  unaffected (the retry loop is not entered).
- ``kMaxAttempts`` is set to 2 (= one retry); empirically the second
  attempt always succeeds on the test environment. Could be raised
  later if reports show single retry is insufficient on other routers.

## Test plan

- [x] Build + flash on M5Stack CoreS3 stackchan board
- [x] NVS reset → captive portal → first submission succeeds
- [ ] Verify on other router models (left to maintainer / community)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,15 @@ documented-only.
   element-type-aware constructor. Contributed via
   [PR #195](https://github.com/kisaragi-mochi/stackchan-mcp/pull/195).
 
+- Fixed: WiFi association now retries with a brief delay and cancels
+  the in-flight `esp_wifi_connect()` on attempt timeout (via
+  `esp_wifi_disconnect()` on the `bits == 0` path), so access points
+  that respond slowly or drop the first 802.11 Association Comeback
+  frame no longer cause the device to fail WiFi at boot. The retry
+  log message also distinguishes timeout vs. driver-reported failure
+  to ease real slow-AP debugging. Contributed via
+  [PR #186](https://github.com/kisaragi-mochi/stackchan-mcp/pull/186).
+
 ### Gateway
 
 - Added: MCP tool surface for the firmware-side Grove Port A generic
--- a/firmware/components/78__esp-wifi-connect/wifi_configuration_ap.cc
+++ b/firmware/components/78__esp-wifi-connect/wifi_configuration_ap.cc
@@ -948,9 +948,22 @@ bool WifiConfigurationAp::ConnectToWifi(const std::string &ssid, const std::stri
         if (bits & WIFI_CONNECTED_BIT) {
             connected = true;
         } else {
+            const bool timed_out = (bits == 0);
+            if (timed_out) {
+                // Timeout — neither WIFI_CONNECTED_BIT nor WIFI_FAIL_BIT
+                // was set, so WIFI_EVENT_STA_DISCONNECTED has not fired
+                // and the driver may still be in `connecting` state.
+                // Cancel the in-flight attempt explicitly before the
+                // retry delay; without this the next esp_wifi_connect()
+                // can return ESP_ERR_WIFI_STATE on a connecting-state
+                // driver (per esp_wifi.h attention 3), making the retry
+                // a no-op on slow / event-dropping APs.
+                esp_wifi_disconnect();
+            }
             ESP_LOGW(TAG,
-                "Attempt %d/%d failed%s",
+                "Attempt %d/%d %s%s",
                 attempt, kMaxAttempts,
+                timed_out ? "timed out (driver may still be connecting)" : "failed",
                 attempt < kMaxAttempts ? " — will retry" : "");
         }
     }
--- a/firmware/components/78__esp-wifi-connect/wifi_configuration_ap.cc
+++ b/firmware/components/78__esp-wifi-connect/wifi_configuration_ap.cc
@@ -871,45 +871,99 @@ bool WifiConfigurationAp::ConnectToWifi(const std::string &ssid, const std::stri
     }
     
     is_connecting_ = true;
-    esp_wifi_scan_stop();
-    xEventGroupClearBits(event_group_, WIFI_CONNECTED_BIT | WIFI_FAIL_BIT);
 
-    wifi_config_t wifi_config;
-    bzero(&wifi_config, sizeof(wifi_config));
-    strlcpy((char *)wifi_config.sta.ssid, ssid.c_str(), 32);
-    strlcpy((char *)wifi_config.sta.password, password.c_str(), 64);
-    wifi_config.sta.scan_method = WIFI_ALL_CHANNEL_SCAN;
-    wifi_config.sta.failure_retry_cnt = 1;
-    
-    ESP_ERROR_CHECK(esp_wifi_set_config(WIFI_IF_STA, &wifi_config));
-    auto ret = esp_wifi_connect();
-    if (ret != ESP_OK) {
-        ESP_LOGE(TAG, "Failed to connect to WiFi: %d", ret);
-        is_connecting_ = false;
-        return false;
-    }
-    ESP_LOGI(TAG, "Connecting to WiFi %s", ssid.c_str());
+    // Upper-level retry loop with delay between attempts.
+    //
+    // Background: in APSTA mode the captive-portal session is on the AP
+    // beacon channel (typically 1), and the target home AP is on some
+    // other channel (e.g. 10). When ConnectToWifi triggers, esp-wifi
+    // performs a Channel Switch Announcement to move both AP and STA to
+    // the home AP's channel, then immediately issues an association
+    // request. The home AP frequently responds with "Association
+    // Response status=30 (Refused Temporarily)" + a Comeback Time in
+    // TUs (= ~1.1s for Buffalo routers, observed) because its own
+    // state hasn't settled yet for the new station.
+    //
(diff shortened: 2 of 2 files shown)


kisaragi-mochi (commented):
Thanks for the thorough investigation. The 802.11 Comeback Time analysis is well-grounded, the before/after log diff is concrete, and the inline comment block makes the rationale easy for future maintainers to follow. Verified locally that the diff is scoped to a single function and the fix logic is sound for the explicit-disconnect failure path.

One thing I'd like to flag from a local review pass:

### Timeout branch may not cancel the in-flight connect

`xEventGroupWaitBits()` can return via three paths:

1. `WIFI_CONNECTED_BIT` set → success.
2. `WIFI_FAIL_BIT` set (from `WIFI_EVENT_STA_DISCONNECTED`) — driver has settled to `disconnected`.
3. Timeout (10s / 25s) → `bits == 0`, and the driver may still be in the `connecting` state.

In case (3), the next `esp_wifi_connect()` can return a state error (e.g. `ESP_ERR_WIFI_STATE`) because the driver is mid-connect, which would prevent the retry from actually running on slow / no-event APs.

A small mitigation:

```cpp
} else {
    if (bits == 0) {
        // Timeout — driver may still be in `connecting` state.
        // Cancel the in-flight attempt before the retry delay so the
        // next esp_wifi_connect() can run cleanly.
        esp_wifi_disconnect();
    }
    ESP_LOGW(TAG, ...);
}
```

Happy to either wait on a follow-up commit from you, or land a small follow-up PR on top after this merges — whichever you prefer.

### Upstream sync

`firmware/components/78__esp-wifi-connect/` is a vendored copy of [78/esp-wifi-connect](https://github.com/78/esp-wifi-connect) (managed via ESP Component Manager, currently `~3.1.3`). The fix would also be valuable upstream so the broader xiaozhi-esp32 ecosystem benefits.

If it would save you time, we're happy to mirror this to an upstream PR on your behalf once it lands here — we'd credit you as the original author (commit `--author` set to you, plus a "Originally contributed by @maha0525 in kisaragi-mochi/stackchan-mcp#186" line in the PR description). If you'd rather send it upstream yourself, please go ahead — just let us know either way and we'll skip our side.

Thanks again for the careful work on this.

maha0525:
Thanks for the careful review!

## Timeout-branch fix (follow-up commit 47f09ac)

You're right — I had only handled `WIFI_FAIL_BIT` implicitly via the `else` branch and didn't differentiate the `bits == 0` timeout case. Per `esp_wifi.h` attention 3 on `esp_wifi_connect()`, calling it while the driver is mid-connect can return `ESP_ERR_WIFI_STATE`, which would make the retry a no-op on slow / event-dropping APs.

The follow-up calls `esp_wifi_disconnect()` only when `bits == 0`, and splits the `ESP_LOGW` message so the operator can tell a timeout from a disconnect-fail:

```cpp
if (bits & WIFI_CONNECTED_BIT) {
    connected = true;
} else {
    const bool timed_out = (bits == 0);
    if (timed_out) {
        // Cancel the in-flight attempt explicitly so the next
        // esp_wifi_connect() doesn't get ESP_ERR_WIFI_STATE.
        esp_wifi_disconnect();
    }
    ESP_LOGW(TAG, "Attempt %d/%d %s%s", ...);
}
```

Local build (ESP-IDF v5.5.4, M5Stack CoreS3 / stackchan target) passes. I haven't been able to reproduce the timeout path on my Buffalo AP — it always returns via `WIFI_FAIL_BIT` — so this part is unverified against a real slow-AP environment. Happy to add follow-ups if you hit the timeout path on yours.

## Upstream sync (78/esp-wifi-connect)

Please go ahead with the mirror on your side once this lands — that's the kindest offer, gratefully accepted. Crediting via `--author` + the "Originally contributed by @maha0525" line works perfectly for me.

kisaragi-mochi:
Just checking in on this one — no rush. Happy to wait at your pace for a timeout-branch update with the `esp_wifi_disconnect()` cleanup.

If you'd prefer to land the current diff as-is and have a follow-up patch from this side cover the timeout-branch cleanup (and optionally mirror it upstream to 78/esp-wifi-connect with credit to you as original author, per the offer in the review), just say the word and we'll move accordingly. Either path works on our side.

kisaragi-mochi:
Apologies for the timing — we crossed paths. By the time the "checking in" comment above was posted, your follow-up commit `47f09ac` and your reply had already landed; we should have refreshed the PR view before commenting.

## Follow-up commit review (47f09ac)

- The `bits == 0` timeout path is correctly distinguished and `esp_wifi_disconnect()` is called only on that branch, matching the `esp_wifi.h` attention 3 contract. The cancel happens before the retry delay so the next `esp_wifi_connect()` runs against a clean driver state.
- The `ESP_LOGW` split (`timed_out ? "timed out (driver may still be connecting)" : "failed"`) surfaces the timeout-vs-disconnect distinction in the operator log — easier to debug a real slow-AP environment in the field. Nice touch.
- Inline rationale referencing `esp_wifi.h` attention 3 keeps the intent explicit for future readers.

CI is green on this branch.

## CHANGELOG entry (heads-up)

A PR-level CHANGELOG check job was added in [PR #203](https://github.com/kisaragi-mochi/stackchan-mcp/pull/203) earlier today, **after your last push**, so this PR now needs a one-line entry under `[Unreleased] > ### Firmware` in `CHANGELOG.md` to pass CI. To spare you another round-trip, we'll add the entry directly to this branch via "Allow edits by maintainers" — apologies for the gate landing mid-PR. If you'd rather add it yourself, just say the word and we'll back off and let you handle it.

## Upstream mirror to `78/esp-wifi-connect`

Gratefully noted on the credit format (`--author` + the "Originally contributed by @maha0525" line). We'll go through the upstream repo's contribution rules first (CONTRIBUTING / PR template / commit convention) before posting anything, and likely open a short heads-up comment there before the PR itself to confirm the convention matches what we have in mind. Will share the upstream PR URL here once it's filed.

Thanks again for the careful work.

kisaragi-mochi:
Upstream mirror filed: [78/esp-wifi-connect#35](https://github.com/78/esp-wifi-connect/pull/35)

- 1 commit, 1 file (`wifi_configuration_ap.cc`), +97/-30
- Author preserved as `Maha <mahomu.tript013@gmail.com>` via `--author`, matching the merged commit metadata here
- "Originally contributed by @maha0525 in kisaragi-mochi/stackchan-mcp#186" line in both the commit message and the PR description
- Surfaced the `--author` email choice transparently in the PR description; if you'd prefer the upstream record use a privacy-safer `users.noreply.github.com` form instead, just say the word and we'll re-author and force-push that branch
- Will keep you posted here if the upstream maintainer requests any changes (commit shape, line wrapping, license header, etc.) — please feel free to drop into that PR thread directly any time

Thanks again for the WiFi-association fix.

Issue #191, opened 2026-05-19
enhancement(firmware): fail-fast on invalid server hello (avoid 10s timeout per malformed candidate)


kisaragi-mochi:
## Symptom

Surfaced in the Codex adversarial review of #187 (Round 5). After #187, the firmware rejects a server hello whose `session_id` is missing, empty, or non-string by returning early from `ParseServerHello` without setting `WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT`. The candidate loop in `OpenAudioChannelInternal` then waits the full event-bit timeout (default 10 seconds) before treating the candidate as failed and falling back to the next URL.

This delays user-visible reconnect by ~10 s per malformed-hello candidate, hiding a protocol contract violation behind a network-timeout path.

## Reproduction (theoretical)

1. Device connects to a gateway candidate URL.
2. Server sends `{"type": "hello", "transport": "websocket"}` without a `session_id` (legacy gateway, version-skewed implementation, or a malformed response).
3. `ParseServerHello` detects the missing session_id, logs `Server hello missing or empty session_id; rejecting candidate`, and returns early without setting the event bit.
4. The main task remains blocked in `xEventGroupWaitBits(WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT, ..., pdMS_TO_TICKS(10000))` for the full 10 s.
5. Timeout elapses; the candidate is reset and the loop proceeds to the next candidate URL (or returns failure if exhausted).

## Why this was not fixed in #187

Carve-out on time-and-overkill grounds:

- The current `stackchan-mcp` gateway always emits a UUID `session_id` in HelloResponse, so the malformed-hello case does not arise on the canonical happy path.
- #187 already converged after five rounds of adversarial review; adding this fix would require a sixth round.
- The failure mode is a ~10 s delay before fallback, **not** a stuck state — the device eventually recovers via candidate fallback or reconnect.

## Suggested fix direction

Introduce a dedicated `WEBSOCKET_PROTOCOL_SERVER_HELLO_FAILED` event bit:

1. Define a new event bit alongside `WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT`.
2. Make `ParseServerHello` set `WEBSOCKET_PROTOCOL_SERVER_HELLO_FAILED` before returning early on missing transport / missing session_id / unsupported transport.
[...]

Issue #190, opened 2026-05-19
feat(firmware/gateway): audio-operation-level identifier to distinguish current operation from same-session stale messages


kisaragi-mochi:
## Symptom

Companion limitation surfaced during Codex adversarial review of #187. After #187 merges, the firmware's tts/listen drop gate accepts any message whose `session_id` matches the current WebSocket-connection `session_id_`. This intentionally allows gateway-driven `say()` / `listen()` MCP control to flow through.

However, **same-session delayed `tts.start` messages can still resurrect `kDeviceStateSpeaking` after a user-initiated `CloseAudioChannel()`**:

1. Gateway calls `say()` → sends `{"type":"tts","state":"start","session_id":"<uuid>"}` (current session).
2. Network delay holds the message in flight.
3. Before the message arrives, the user touches the head → `CloseAudioChannel()` → `audio_channel_open_=false` (WebSocket and `session_id_` are kept alive per #136).
4. Delayed `tts.start` arrives, `session_id` matches `session_id_`, message passes the #187 gate.
5. `Application::OnIncomingJson` schedules `kDeviceStateSpeaking` → the device begins to speak after the user's close gesture.

This recreates a variant of the #136 stale-frame lifecycle issue (stale `tts.start` resurrects `kDeviceStateSpeaking`), but on the **same-WebSocket-connection** axis rather than the cross-session axis that #136 / #187 addressed.

## Why this is not fixed in #187

The session_id-based gate in #187 distinguishes "current WebSocket session" from "stale WebSocket session". It does not distinguish "current audio operation" from "stale audio operation within the same WebSocket session". A finer-grained mechanism is required.

## Suggested fix direction

Add an **audio-operation-level identifier** that the gateway includes in every `tts.*` / `listen.*` message (separate from the WebSocket `session_id`):

- **Option A — operation_id token**: gateway generates a fresh `operation_id` for each `say()` / `listen()` invocation. Firmware tracks the latest accepted `operation_id` and drops messages with a different one.
- **Option B — firmware-side audio session-armed flag**: firmware maintains a generation counter incremented at `CloseAudioChannel()`. Gateway includes the current generation in `tts.start` / `listen.start`; firmware advances the counter on `CloseAudioChannel` or operation completion. Stale messages carrying an older generation are dropped.

Both options require a firmware + gateway change. Option A is simpler (gateway is the authority); Option B has stronger invariants if the gateway is untrusted (not currently a concern for stackchan-mcp).

[...]

Issue #189, opened 2026-05-19
bug(firmware): hello-then-disconnect race in OpenAudioChannelInternal cancels reconnect timer


kisaragi-mochi:
## Symptom

After PR #136 lands, in the `WebsocketProtocol::OpenAudioChannelInternal()` main task path, the sequence after `xEventGroupWaitBits()` returns from a `WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT` includes (around `firmware/main/protocols/websocket_protocol.cc:441-444`):

```cpp
// mirror notify_disconnect token to the class member
// clear intentional_close_
// reset reconnect_interval_ms_
// StopReconnectTimer()
return true;
```

If a server-side close arrives **between** `ParseServerHello()` setting the event bit and the main task resuming, the per-socket `OnDisconnected` lambda fires on the WS task, runs `ScheduleReconnect()`, and arms a reconnect timer. The main task then resumes from `xEventGroupWaitBits()` and unconditionally calls `StopReconnectTimer()`, which cancels the just-scheduled reconnect.

`OpenAudioChannelInternal(false, false)` returns `true` to the reconnect timer caller (which originally invoked it), so no failure-reschedule path is taken. The transport ends up disconnected with no reconnect scheduled. The device sits in `State: listening → idle` with the WebSocket gone, and MCP control is unavailable until the next user action (touch, wake word).

## Reproduction (theoretical)

1. Reconnect timer fires → `OpenAudioChannelInternal(false, false)` runs on the reconnect-timer task.
2. New socket connects, hello sent.
3. WS task receives server hello, runs `ParseServerHello()`, sets the event bit.
4. Between (3) and (5): a server-side TCP RST or close arrives. WS task's `OnDisconnected` lambda fires, invokes `ScheduleReconnect()`, which posts a new reconnect timer.
5. Reconnect-timer task wakes from `xEventGroupWaitBits()`, mirrors the now-armed `notify_disconnect`, clears `intentional_close_`, resets `reconnect_interval_ms_`, calls `StopReconnectTimer()` — **cancelling the timer just scheduled in step 4**.
6. Function returns `true` to the reconnect-timer caller, which treats the attempt as successful and does not reschedule.

Net effect: transport is dead, no reconnect is scheduled, device requires user action to recover.

## Real-world probability

[...]

Issue #188, opened 2026-05-19
bug(firmware): OnDisconnected writes audio_channel_open_ before notify_disconnect guard — stale callback can corrupt fresh session


kisaragi-mochi:
## Symptom

After PR #136 lands, `WebsocketProtocol::OnDisconnected` (`firmware/main/protocols/websocket_protocol.cc:384-395`) writes `audio_channel_open_.store(false)` **before** checking the per-socket `notify_disconnect` token. The token guard runs immediately after, on lines 393-396, and only there does the lambda return early when the token is disarmed.

When `OpenAudioChannelInternal()` performs the standard "reset old socket then create new socket" sequence, the old socket's `OnDisconnected` callback can fire late (after `websocket_.reset()` returns) via the ESP-IDF WebSocket client's event loop. The token is correctly disarmed before the new socket starts, so the disconnect handler returns early — but only after it has already written `audio_channel_open_=false`.

If the new socket has already completed `ParseServerHello(... arm_audio_channel=true)` by the time the late callback fires, the global `audio_channel_open_` is silently flipped to `false` and the fresh audio session is corrupted from the caller's perspective (`IsAudioChannelOpened()` now returns `false`, `CanEnterSleepMode()` blocks sleep — both inverted from intent).

## Reproduction (theoretical)

1. User opens audio (`OpenAudioChannel()`), session A established, `audio_channel_open_=true`, socket A's `notify_disconnect=armed`.
2. User triggers a re-open path (touch toggle, or any caller that hits `OpenAudioChannelInternal`).
3. `OpenAudioChannelInternal` disarms socket A's `notify_disconnect` (the shared `std::atomic<bool>`), resets `websocket_`, creates socket B with its own fresh `notify_disconnect=false`.
4. Socket B handshake completes, `ParseServerHello(... arm_audio_channel=true)` runs, `audio_channel_open_.store(true)`.
5. Socket A's `OnDisconnected` callback finally fires (delayed via the ESP-IDF event loop).
6. Line 385: `audio_channel_open_.store(false)` runs unconditionally.
7. Lines 393-396: the disarmed `notify_disconnect` causes early return — but only **after** step 6.

Net effect: callers see `IsAudioChannelOpened()=false` despite the live socket B session being arm=true.

## Real-world probability

[...]

PR #192, opened 2026-05-19
fix(firmware): #187 session_id-based gating for tts/listen messages

### LINKED ISSUE #187: bug(firmware): gateway-driven tts/listen messages silently dropped after CloseAudioChannel (companion to #136)
## Symptom

Companion to #136. After #136 lands, the firmware keeps the WebSocket alive across `CloseAudioChannel()` (the original intent: MCP control should survive an audio-session close). However, the drop logic introduced in the same PR (`firmware/main/protocols/websocket_protocol.cc:363-370`) discards all inbound `tts.*` / `listen.*` JSON messages whenever `audio_channel_open_=false`.

The gateway's `say()` and `listen()` MCP tools rely on sending `{"type":"tts","state":"start"}` and `{"type":"listen","state":"start"}` over the same WebSocket to drive the firmware state machine. Without those notifications, the device firmware does not enter `kDeviceStateSpeaking` / listening mode, and subsequent binary Opus frames are also discarded by the existing `OnIncomingAudio` gate.

**Net effect**: after a user closes the audio channel (e.g. touches the head), `say()` / `listen()` MCP tools report success but no audio plays / no microphone input is captured. Silent failure.

## Reproduction (post-#136)

1. Device connects, OpenAudioChannel runs normally, `audio_channel_open_=true`.
2. User touches the head → `CloseAudioChannel()` → `audio_channel_open_=false`, WebSocket stays alive (this is #136's intended behaviour).
3. Gateway-side agent calls the `say("hello")` MCP tool.
4. Gateway sends `{"type":"tts","state":"start","session_id":"<uuid>"}` followed by Opus frames.
5. Firmware drop logic (`websocket_protocol.cc:363-370`) discards the `tts.start` JSON because `audio_channel_open_=false`.
6. `kDeviceStateSpeaking` never entered; binary frames are then also discarded by `OnIncomingAudio`'s state-machine gate.
7. `say()` returns success with `sent_frames=N>0` but no audio plays.

## Root cause

#136's stale-frame protection conflates "logically closed audio channel" with "discard all audio-session messages". The gateway-driven design intentionally sends `tts.start` / `listen.start` over the preserved transport while the channel is logically closed — that is the entire point of keeping the WebSocket alive.

The firmware already tracks `session_id_` per WebSocket handshake (`websocket_protocol.cc`, populated from server hello at `ParseServerHello()`), and the gateway already echoes its UUID-based `session_id` in every `tts.*` / `listen.*` message (`gateway/stackchan_mcp/esp32_client.py:237, 262`). The firmware does not currently validate these inbound `session_id` values, so it cannot distinguish a current-session message from a stale one.

## Fix direction

Replace the binary `audio_channel_open_` predicate in the drop logic with a session_id-based gate:

1. **Retain `session_id_` across `CloseAudioChannel()`** (currently cleared at `websocket_protocol.cc:188`). The session_id is owned by the WebSocket connection lifetime, not the audio-session lifetime, and the gateway echoes the same value across its full connection.

2. **Replace the drop predicate at `websocket_protocol.cc:363-370`** with: drop only when the incoming `session_id` is missing or does not match `session_id_`. Messages whose `session_id` matches `session_id_` are passed to `on_incoming_json_`. This means:
   - Gateway-current-session `tts.start` / `listen.start` → passes through, drives the state machine.
   - Stale `tts.start` from a previous (different) session_id → still dropped (the original Finding 2 protection still holds).

3. **Audit the binary frame drop guard** at `websocket_protocol.cc:315-321`. The existing `OnIncomingAudio` gate (`kDeviceStateSpeaking`-only) likely makes the new guard redundant; if so, remove it to keep behaviour symmetric with the JSON path.

## Acceptance criteria

- After the user touches the head (`CloseAudioChannel()` → `audio_channel_open_=false`), the gateway `say(text)` MCP tool produces audible output on the device.
- After the same close, the gateway `listen()` MCP tool captures microphone input from the device.
- Stale `tts.start` carrying a session_id that does not match the current `session_id_` is still dropped — verified by a firmware-side log or a manual injection test.
- Unit tests confirm session_id check behaviour (mismatch dropped, match accepted), and the existing gateway TTS / STT tests continue to pass.

## Refs

- Companion PR: #136 (`fix: keep WebSocket alive when CloseAudioChannel is called`)
- Gateway TTS pipeline: `gateway/stackchan_mcp/tts/orchestrator.py:184-218`, `gateway/stackchan_mcp/esp32_client.py:222-268`
- Firmware `session_id_` storage: `firmware/main/protocols/protocol.h:89`, set in `firmware/main/protocols/websocket_protocol.cc:544-547` (ParseServerHello), cleared at `firmware/main/protocols/websocket_protocol.cc:188` (current behaviour — to be changed)
- Firmware tts/listen handler: `firmware/main/application.cc:515-588`

kisaragi-mochi:
## Summary

Companion fix to #136 (merged as `41a22d2`). #136 keeps the WebSocket alive across `CloseAudioChannel()` so MCP control survives an audio-session close. However, the drop logic introduced in #136 (`firmware/main/protocols/websocket_protocol.cc:363-370`) discarded all inbound `tts.*` / `listen.*` JSON messages whenever `audio_channel_open_=false`, which silently broke the gateway's `say()` and `listen()` MCP tools — they send `{"type":"tts","state":"start"}` / `{"type":"listen","state":"start"}` over the same WebSocket to drive the firmware state machine.

This PR replaces the binary `audio_channel_open_` predicate with a session_id-based gate: gateway-current-session messages pass through, stale (mismatched session_id) messages are still dropped.

Closes #187.

## Why session_id-based gating

The gateway already generates a per-connection UUID-based `session_id` and echoes it in every `tts.*` / `listen.*` message (`gateway/stackchan_mcp/esp32_client.py:419, 237, 262`). The firmware already tracks `session_id_` per WebSocket handshake (`firmware/main/protocols/websocket_protocol.cc` via `ParseServerHello`). The fix validates these against each other on the inbound side, which:

- **Preserves the original stale-frame protection from #136**: a `tts.start` carrying a different session_id (a previous WebSocket session) is still dropped.
- **Allows the gateway-driven audio control path to flow through**: current-session `tts.start` / `listen.start` reach the firmware state machine, so `say()` and `listen()` MCP tools work after a user-initiated `CloseAudioChannel()`.
- **Confined to one file** plus public docs alignment.

## Changes

`firmware/main/protocols/websocket_protocol.cc`:

- **`CloseAudioChannel`**: retain `session_id_` across the close. The session_id is owned by the WebSocket connection lifetime, not the audio-session lifetime — the gateway echoes the same UUID across its full connection, and `ParseServerHello` overwrites `session_id_` on reconnect or a fresh hello exchange.
- **`OpenAudioChannelInternal`**: clear `session_id_` at the start of a fresh socket attempt so a malformed or version-skewed server hello cannot leave the previous connection's id active as the tts/listen gate key.
- **`ParseServerHello`**: require a non-empty string `session_id` in the server hello. A missing, empty, or non-string `session_id` is rejected (no event bit, no `notify_disconnect` arm) so the candidate loop falls back to the next URL or surfaces failure.
- **`OnData` JSON branch**: replace the `audio_channel_open_`-based drop predicate with a `session_id` mismatch check. Current-session messages pass through; missing / mismatched session_id messages are still logged and dropped.
- **`OnData` binary branch**: replace the `audio_channel_open_` guard with an early `kDeviceStateSpeaking` gate via `Application::GetInstance().GetDeviceState()`. Avoids BinaryProtocol2/3 header parsing and AudioStreamPacket payload allocation on the closed-channel path.

`firmware/docs/websocket.md`, `firmware/docs/websocket_zh.md`:

- Tighten the public WebSocket protocol contract: `session_id` in the server hello is now documented as a **required non-empty string** (previously documented as optional), matching the new firmware-side enforcement above.

## Known limitations (carved-out follow-ups)

These were surfaced during five rounds of Codex adversarial review on this PR. They do not block #187 and are tracked separately:

- **#190 — Same-session delayed `tts.start` after `CloseAudioChannel`** (low real-world probability, recoverable). A `tts.start` already in flight from the gateway when the user closes the audio channel will pass the session_id gate. Resolving this requires an **audio-operation-level identifier** (separate from the WebSocket session_id) so the firmware can distinguish "current audio operation" from "stale audio operation within the same WebSocket session". Two design options proposed in the Issue.
- **#191 — Invalid server hello burns the full 10s candidate timeout before fallback** (low real-world probability, ~10s startup delay only). The `ParseServerHello` rejection returns early without setting the event bit, so the candidate loop waits the full timeout before treating the candidate as failed. The canonical `stackchan-mcp` gateway always sends a valid `session_id`, so this only surfaces with legacy / malformed gateways. Fix direction: dedicated `SERVER_HELLO_FAILED` event bit for fail-fast paths.

## Compatibility with #136

This PR replaces the gate introduced in #136 with the session_id-based gate. It is intended to land soon after #136 (which has just merged as `41a22d2`) to restore gateway-driven `say()` / `listen()` semantics.

## Validation

- **Build**: confirmed via `python ./scripts/release.py stackchan` (Docker `espressif/idf:v5.5.2`). Artifacts produced: `build/xiaozhi.bin`, `build/merged-binary.bin`, `releases/v2.2.6_stackchan.zip`.
- **Gateway tests**: `cd gateway && uv run pytest` — all existing tests pass (255 passed, 1 skipped). Gateway-side behaviour is unchanged.
- **Code review**: five rounds of Codex adversarial-review run before opening this PR. Findings addressed in-PR:
  - **Finding 2 (Round 1)** — binary frames parsed before drop, potential out-of-bounds via untrusted `payload_size`. Addressed: early `kDeviceStateSpeaking` gate replaces the `audio_channel_open_` guard.
  - **Finding 3 (Round 1)** — malformed server hello leaves stale `session_id` active. Addressed: `OpenAudioChannelInternal` clears `session_id_` at fresh-socket-attempt time; `ParseServerHello` requires non-empty `session_id`.
  - **Finding 4 (Round 4)** — public docs contract stale. Addressed: `websocket.md` / `websocket_zh.md` updated to require non-empty `session_id`.
  - **Finding 1 (Round 1)** carved out to **#190**.
  - **Finding 5 (Round 5)** carved out to **#191**.

## Test plan (real-device, post-merge)

Pending real-device confirmation on M5Stack CoreS3 stackchan board:

- [ ] After user touch closes the audio channel (`audio_channel_open_=false`, WS alive per #136), `gateway say("hello")` produces audible output on the device.
- [ ] After the same close, `gateway listen()` captures audio from the device microphone.
- [ ] Injecting a `tts.start` with a mismatched `session_id` (e.g. from a stale separate connection) is still dropped — verified via firmware serial log (`Dropping tts message (session_id mismatch or missing)`).
- [ ] `CanEnterSleepMode()` continues to return false when audio session is open and true when closed — `audio_channel_open_` semantic preserved (this PR does not touch the flag itself).

## Refs

- Closes #187
- Companion to #136 (merged as `41a22d2`)
- Carved-out follow-ups: #190 (audio-operation-level identifier), #191 (fail-fast on invalid hello)
- Related lifecycle Issues (race conditions, low real-world probability, independent of #187): #188 (stale disconnect callback), #189 (hello-then-disconnect timer cancel)

### DIFF
--- a/firmware/docs/websocket.md
+++ b/firmware/docs/websocket.md
@@ -43,7 +43,7 @@ This document describes the WebSocket communication protocol between the device 
 
 4. **Server replies with "hello"**
    - The device waits for a JSON message whose `"type"` is `"hello"` and whose `"transport"` is `"websocket"`.
-   - The server may include a `session_id`; the device will store it.
+   - The server **must** include a non-empty string `session_id`; the device stores it and uses it to validate subsequent `tts.*` / `listen.*` messages. A hello without a valid `session_id` is rejected and the device falls back to the next candidate URL (or surfaces a connection failure when no candidates remain).
    - Example:
    ```json
    {
@@ -232,9 +232,9 @@ WebSocket text frames carry JSON. The most common `"type"` values and their sema
 1. **Hello**
    - The handshake acknowledgement.
    - Must include `"type": "hello"` and `"transport": "websocket"`.
+   - Must include a non-empty string `session_id` which the device records; subsequent `tts.*` / `listen.*` messages must echo the same value to pass the firmware's session-match gate.
    - May include `audio_params`, meaning the audio parameters the server expects / the canonical set agreed with the device.
-   - May include a `session_id` which the device records.
-   - Once received, the device sets the "audio channel open" event.
+   - Once received (and `session_id` validated), the device sets the "audio channel open" event.
 
 2. **STT**
    - `{"session_id": "xxx", "type": "stt", "text": "..."}`
--- a/firmware/docs/websocket_zh.md
+++ b/firmware/docs/websocket_zh.md
@@ -42,7 +42,7 @@
 
 4. **服务器回复 "hello"**  
    - 设备等待服务器返回一条包含 `"type": "hello"` 的 JSON 消息，并检查 `"transport": "websocket"` 是否匹配。  
-   - 服务器可选下发 `session_id` 字段，设备端收到后会自动记录。  
+   - 服务器**必须**下发非空字符串 `session_id` 字段，设备端会记录并用于后续 `tts.*` / `listen.*` 消息的会话匹配校验。缺失、空或非字符串的 `session_id` 会导致设备拒绝该握手并回退到下一个候选 URL（或在候选用尽时上报连接失败）。  
    - 示例：
    ```json
    {
@@ -221,9 +221,9 @@ WebSocket 文本帧以 JSON 方式传输，以下为常见的 `"type"` 字段及
 1. **Hello**  
    - 服务器端返回的握手确认消息。  
    - 必须包含 `"type": "hello"` 和 `"transport": "websocket"`。  
+   - 必须包含非空字符串 `session_id` 字段，设备端会记录；后续的 `tts.*` / `listen.*` 消息必须携带相同的 `session_id` 才能通过固件侧的会话匹配校验。  
    - 可能会带有 `audio_params`，表示服务器期望的音频参数，或与设备端对齐的配置。   
-   - 服务器可选下发 `session_id` 字段，设备端收到后会自动记录。  
-   - 成功接收后设备端会设置事件标志，表示 WebSocket 通道就绪。
+   - 成功接收（且 `session_id` 通过校验）后设备端会设置事件标志，表示 WebSocket 通道就绪。
 
 2. **STT**  
    - `{"session_id": "xxx", "type": "stt", "text": "..."}`
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -161,6 +161,7 @@ void WebsocketProtocol::CloseAudioChannel(bool send_goodbye) {
     // the app transitions back to idle while the WebSocket stays connected
     // for continued MCP control.
     audio_channel_open_.store(false);
+    // Keep session_id_ across a logical audio-session close; it belongs to the WebSocket connection lifetime.
     ESP_LOGI(TAG, "CloseAudioChannel: keeping WebSocket alive for MCP");
     if (on_audio_channel_closed_ != nullptr) {
         on_audio_channel_closed_();
@@ -185,6 +186,11 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_aud
     }
     StopReconnectTimer();
     websocket_.reset();
+    // Clear session_id_ at the start of a fresh socket attempt so a
+    // malformed or version-skewed server hello cannot leave the previous
+    // connection's id active as the tts/listen gate key. CloseAudioChannel
+    // intentionally keeps session_id_ alive across a logical audio-session
+    // close while the WebSocket stays connected.
     session_id_ = "";
     xEventGroupClearBits(event_group_handle_, WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT);
 
@@ -313,10 +319,10 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_aud
 
         websocket_->OnData([this, notify_disconnect, arm_audio_channel](const char* data, size_t len, bool binary) {
             if (binary) {
-                // Drop inbound audio when the audio channel is logically
-                // closed. Without this guard, a late TTS frame from the
-                // previous session could resurrect kDeviceStateSpeaking.
-                if (!audio_channel_open_.load()) {
+                // Drop binary frames before parsing when the device is not
+                // speaking. This mirrors Application::OnIncomingAudio's
+                // downstream gate while avoiding stale-frame parse/allocation.
+                if (Application::GetInstance().GetDeviceState() != kDeviceStateSpeaking) {
                     return;
                 }
                 if (on_incoming_audio_ != nullptr) {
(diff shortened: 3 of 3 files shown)


kisaragi-mochi:
## Real-device verification report (post-merge)

Verified on M5Stack CoreS3 with firmware HEAD `1e4e123` + PR #192 commit `9be82e4` (`xiaozhi.bin` flashed to `0x20000`).

### Status summary

| # | Test plan item | Status | Notes |
|---|---|---|---|
| 1 | `say()` after touch-close | △ baseline ✅ | See "Touch-close trigger investigation" below for the literal-condition gap. |
| 2 | `listen()` after touch-close | △ baseline ✅ | Same as Item 1. |
| 3 | Stale session_id `tts.start` dropped | **✅ unintentionally observed** | See "Item 3 — mismatch drop verified live" below. |
| 4 | `CanEnterSleepMode()` regression | ✅ code verify | PR body confirms `audio_channel_open_` flag itself is unchanged by this PR. |

### Baseline verification (current-session pass-through)

The current-session pass-through path that PR #192 preserves was confirmed via baseline `say()` / `listen()` invocations immediately after the firmware-side session_id (`22c65bfe-...`) was issued at boot:

- `say(text="こんにちは")` returned `frame_count=18, duration_ms=1080` and the device produced audible output.
- `listen(duration_ms=5000, language="ja")` with silence input returned `frame_count=84, duration_ms=5040, text=""`. 84 frames × 60 ms = 5040 ms — full capture window completed.

Both confirm that session_id-matching `tts.*` / `listen.*` messages flow through the new gate as intended.

### Item 3 — mismatch drop verified live

While preparing the literal touch-close re-verification, the device hit a 633s `Channel timeout` and the firmware initiated a WebSocket reconnect, issuing a new session_id (`fcc2eca1-...`). The gateway-side `ESP32Connection` had no path to observe the new id and continued to attach the previous session_id (`22c65bfe-...`) on outbound messages.

When `listen(duration_ms=5000)` was invoked shortly after the reconnect, the gateway returned:

```json
{"engine": "faster-whisper", "text": "", "language": "ja",
 "duration_ms": 0, "frame_count": 0, "sample_rate": 16000}
```

The corresponding firmware serial log:

```
E (5869195) Protocol: Channel timeout 633 seconds
I (5869195) StateMachine: State: idle -> connecting
I (5870385) WS: Session ID: fcc2eca1-30e9-4c8b-9418-270d0b180c19
I (5870395) WS: Connected to websocket server candidate 1/1: ...
I (5870395) StateMachine: State: connecting -> listening
I (5874245) StateMachine: State: listening -> idle
```

The immediate `listening -> idle` transition (without any Opus frames being emitted) plus `frame_count=0` on the gateway side confirm that the `listen.start` was dropped by the firmware's PR #192 gate due to session_id mismatch. This is exactly the cross-session-stale-message scenario that PR #192 was designed to drop.

This is the strongest available real-device evidence of PR #192's drop predicate working as designed.

### Touch-close trigger investigation (Items 1 and 2, literal condition)

The test plan's literal condition "After user touch closes the audio channel" turned out to be non-reproducible via the touch handler on this firmware: `firmware/main/boards/stackchan/stackchan.cc` touch handler does not call `CloseAudioChannel()`. All `CloseAudioChannel()` callers live in `firmware/main/application.cc` (lines 291, 748, 994, 1013, 1082, 1136, 1149) and are driven by network disconnect, voice-command-end, sleep, reboot, OTA, or AEC mode change — not by touch directly.

Functionally, however, the relevant property (current-session pass-through across an audio-channel close) is established by combining (a) the baseline pass-through above and (b) the PR's own code change — `audio_channel_open_` is no longer the gate predicate, session_id is. The literal touch-close re-run is deferred unless additional cross-cutting evidence is required.

### Follow-up issue filed

The reconnect-with-new-session_id observation also surfaced a structural gap on the gateway side: `ESP32Connection` does not pick up the new session_id from the firmware's reconnect server hello, so all post-reconnect audio operations are silently dropped until the gateway is restarted. This is a strict superset of the test failure observed above — same root cause, broader impact.

Filed as #194: `bug(gateway): missing session_id re-sync after ESP32 WebSocket reconnect — stale session_id causes all post-reconnect tts/listen to be dropped by #192 gate`. The proposed fix is gateway-side only (no wire protocol change). Scope distinct from #190 (same-WebSocket-connection audio-operation-level identifier) and from #188 / #189 (firmware-side reconnect races); see #194's "Relationship to #190" section for the axis-level differentiation.

### Conclusion

PR #192's firmware-side session_id gate is functioning as designed in both directions:

- Current-session messages flow through (Items 1, 2 baseline).
- Stale session_id messages are dropped (Item 3, real-device evidence).

Real-device verification is complete for the wire-flow properties of this PR. The companion gateway-side reconnect-aware session_id update is tracked in #194 and is required to keep the audio path live across firmware-initiated WebSocket reconnects.

PR #195, opened 2026-05-19
feat(firmware/mcp_server): add Array property type for tools with list arguments

maha0525:
## Summary

Adds `kPropertyTypeArray` (with `PropertyElementType` for Integer / String element types) to the firmware-side MCP server's `Property` type system, enabling MCP tools to declare and accept array arguments directly. Purely additive — no existing tool uses the new type, all non-array `Property` usages compile and behave unchanged.

## Motivation

The existing `PropertyType` (`Boolean` / `Integer` / `String`) forces tools handling ordered byte/list data to encode their input as opaque strings. Downstream tools that work on ordered byte sequences (e.g. generic I2C-bus operations for external M5Stack Unit modules) end up parsing JSON-strings on the firmware side, which is fragile and obscures the schema.

A first-class array type lets the MCP server emit a clean `{"type":"array","items":{...}}` JSON Schema, the client-side framework handles validation, and the firmware-side handler receives a `std::vector<int>` / `std::vector<std::string>` directly.

This is the type-system foundation. A follow-up PR uses it for generic I2C tools that expose Port A devices to the gateway without per-Unit firmware rebuilds.

## Changes

**`firmware/main/mcp_server.h`**:
- Add `kPropertyTypeArray` to `PropertyType` enum
- Add `PropertyElementType` enum (`Integer` / `String`)
- Extend `Property::value_` variant with `std::vector<int>` and `std::vector<std::string>`
- Add `Property` constructors for arrays: element-type only, and element-type with per-element integer range
- `Property::to_json` emits `{"type":"array","items":{...}}` including per-element `minimum` / `maximum` for integer arrays
- `Property::set_value` validates element types and enforces per-element range checks

**`firmware/main/mcp_server.cc`**:
- `DoToolCall` parses cJSON arrays into `std::vector<int>` or `std::vector<std::string>` with element-type validation

## Compatibility

Purely additive: no existing `PropertyType` value or `Property` constructor changes signature, no existing tool uses `kPropertyTypeArray`, schema emission and value handling for existing tools are unchanged.

## Test plan

- [x] `idf.py build` (stackchan target, ESP-IDF v5.5.4): 2125/2125 steps, `xiaozhi.bin` 0x30beb0 bytes, no warnings introduced
- [x] Existing tools (`move_head`, `set_avatar`, `set_blink`, etc.) register and execute as before
- [x] On-device verification of the new Array type via the follow-up PR's `self.i2c.write` / `self.i2c.write_read` tools (integer array arguments end-to-end)

### DIFF
--- a/firmware/main/mcp_server.cc
+++ b/firmware/main/mcp_server.cc
@@ -535,6 +535,35 @@ void McpServer::DoToolCall(int id, const std::string& tool_name, const cJSON* to
                 } else if (argument.type() == kPropertyTypeString && cJSON_IsString(value)) {
                     argument.set_value<std::string>(value->valuestring);
                     found = true;
+                } else if (argument.type() == kPropertyTypeArray && cJSON_IsArray(value)) {
+                    int array_size = cJSON_GetArraySize(value);
+                    if (argument.element_type() == kPropertyElementTypeInteger) {
+                        std::vector<int> int_array;
+                        int_array.reserve(array_size);
+                        for (int i = 0; i < array_size; i++) {
+                            auto item = cJSON_GetArrayItem(value, i);
+                            if (!cJSON_IsNumber(item)) {
+                                throw std::invalid_argument("Array element " + std::to_string(i) +
+                                                            " of '" + argument.name() + "' is not an integer");
+                            }
+                            int_array.push_back(item->valueint);
+                        }
+                        argument.set_value<std::vector<int>>(int_array);
+                        found = true;
+                    } else if (argument.element_type() == kPropertyElementTypeString) {
+                        std::vector<std::string> string_array;
+                        string_array.reserve(array_size);
+                        for (int i = 0; i < array_size; i++) {
+                            auto item = cJSON_GetArrayItem(value, i);
+                            if (!cJSON_IsString(item)) {
+                                throw std::invalid_argument("Array element " + std::to_string(i) +
+                                                            " of '" + argument.name() + "' is not a string");
+                            }
+                            string_array.push_back(item->valuestring);
+                        }
+                        argument.set_value<std::vector<std::string>>(string_array);
+                        found = true;
+                    }
                 }
             }
 
--- a/firmware/main/mcp_server.h
+++ b/firmware/main/mcp_server.h
@@ -52,24 +52,39 @@ using ReturnValue = std::variant<bool, int, std::string, cJSON*, ImageContent*>;
 enum PropertyType {
     kPropertyTypeBoolean,
     kPropertyTypeInteger,
-    kPropertyTypeString
+    kPropertyTypeString,
+    kPropertyTypeArray
+};
+
+// Element type for kPropertyTypeArray properties.
+// Only meaningful when the surrounding Property has type kPropertyTypeArray.
+enum PropertyElementType {
+    kPropertyElementTypeInteger,
+    kPropertyElementTypeString
 };
 
 class Property {
 private:
     std::string name_;
     PropertyType type_;
-    std::variant<bool, int, std::string> value_;
+    std::variant<bool, int, std::string, std::vector<int>, std::vector<std::string>> value_;
     bool has_default_value_;
-    std::optional<int> min_value_;  // 新增：整数最小值
-    std::optional<int> max_value_;  // 新增：整数最大值
+    std::optional<int> min_value_;  // 整数最小值 (kPropertyTypeInteger)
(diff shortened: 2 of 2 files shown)


kisaragi-mochi (commented):
Thanks for the careful work and the pre-PR self-review — this is a clean foundation type-system extension. Verified locally that the array path emits `{"type":"array","items":{...}}` cleanly and the existing Boolean / Integer / String paths are untouched.

A local adversarial review surfaced three improvement candidates. None of them are shipping blockers given the LAN-trusted MCP context this gateway runs in (no untrusted external caller; integer arrays are byte-level for the I2C use case in PR #196), so flagging these as `Suggested` per Conventional Comments — feel free to fold in, punt to a follow-up, or skip at your discretion.

### Suggested (optional improvements)

**(1) Integer array elements accept decimal JSON numbers** (`firmware/main/mcp_server.cc:545-551`)

`cJSON`'s `valueint` truncates any decimal payload to an int before `set_value` runs. A caller sending `[10.9, 200]` ends up with `[10, 200]`; range checks evaluate the truncated value rather than the original.

For the immediate I2C use case (PR #196), `bytes` are 0–255 integers, so callers should always send integers and there's no semantic difference. The risk surfaces if a future array tool encodes decimal-meaningful data (angles, PWM duty, coordinates). A guard could reject non-integral numbers and out-of-int-range values via `valuedouble` + `isfinite` + an exact integer check before pushing into the vector.

**(2) Array payload size is unbounded** (`firmware/main/mcp_server.cc:539-565`)

The handler `reserve`s the client-provided array length without a cap. For a LAN-trusted MCP server this is not an attack vector — but a stray oversized request from an LLM client could still drive ESP32 memory exhaustion before any tool callback runs. Adding `max_items` metadata to array `Property` definitions, emitting `maxItems` in `to_json`, and rejecting oversized arrays pre-`reserve` would give a defense-in-depth layer at low cost. A conservative default cap (e.g. 256, matching the I2C `n_bytes` cap) might be enough.

**(3) The generic default-value constructor admits broken array defaults** (`firmware/main/mcp_server.h:88-91`)

The template default-value constructor accepts `kPropertyTypeArray` with a `std::vector` default, but leaves `element_type_` unset. The result emits `{"type":"array","items":{}}` and any later argument lookup hits `element_type().value()` on an empty optional. No call site is exercising this today; the risk is a future tool author landing in this gotcha. Pure doc + a small fail-fast in `to_json` (or constraining the generic constructor to reject `kPropertyTypeArray`) would close the gap.

---

If you'd like to fold any of these in, take your time — there's no rush. If you'd rather skip and land as-is, we're happy to handle the cleanups as small follow-up commits — just let us know which path is easier for you.

Same review structure used on PR #196 in a separate comment (one item there flagged as `Recommended for landing`).

Thanks again — this is a useful foundation.

maha0525:
Thanks for the careful review! Folded in items 2 and 3 in commit `7720940`:

- **Item 2 (maxItems)**: added optional `max_items` via setter (`Property::set_max_items`), so existing constructors stay unchanged. `set_value` rejects oversized arrays before reserving memory; both `std::vector<int>` and `std::vector<std::string>` payloads are checked.
- **Item 3 (broken default ctor)**: the generic default-value constructor now rejects `kPropertyTypeArray` with a clear error pointing future authors to the element-type-aware `Property(name, type, element_type, ...)` form.

Item 1 (decimal JSON truncation in `cJSON valueint`) intentionally deferred — current I2C-bytes use case in PR #196 is integer-only, easier to revisit when a tool actually encodes decimal-meaningful values and we know the right acceptance shape.

Build verified clean. PR #196 then merges this branch to pick up the `max_items` support and applies it to the `bytes` / `write_bytes` props (256 cap, symmetric with the existing `n_bytes` read cap).

maha0525:
For visibility: filed item 1 (decimal truncation) as #200, with the suggested-fix sketch and deferral rationale included. Happy to revisit when a tool actually encodes decimal-meaningful array values, or if a contributor wants to pick it up as a small follow-up PR.

kisaragi-mochi:
Merged via c6f8a74. Thanks for the careful folding of items 2 and 3, and for filing item 1 as #200 with the suggested-fix sketch — that makes the deferral easy to track.

The Array property type foundation lands cleanly. The optional `set_max_items` setter is a nice design choice — it keeps the existing constructor signatures unchanged while still adding the bounded-allocation guarantee, which makes future array tools easy to extend without breaking older callers. The fail-fast in the generic default-value constructor for `kPropertyTypeArray` also closes the dereference-of-empty-optional gap with a clear message that points the next contributor to the right form.

Looking forward to more contributions.

PR #196, opened 2026-05-19
feat(firmware/stackchan): add Port A I2C bus and expose generic I2C tools

maha0525:
## Summary

Adds a dedicated I2C master bus on the stackchan board's Grove Port A (GPIO 2 SDA / GPIO 1 SCL, I2C controller 0), alongside the existing internal bus (GPIO 12/11, I2C controller 1) used by on-board ICs. Exposes four MCP tools that operate on the Port A bus only, letting users drive attached M5Stack Unit modules (ENV III, ToF, gas sensor, PaHub, etc.) over MCP without per-Unit firmware rebuilds.

**Depends on PR #195** (kPropertyTypeArray) — the new tools' `bytes` / `write_bytes` parameters use the Array property type.

## Motivation

The stackchan board already hosts an MCP server with physical-embodiment tools (head movement, avatar, LEDs, etc.). Adding generic I2C bus operations lets the same server expose external sensors / actuators on Port A, so end users can plug in an M5Stack Unit module and drive it via Python on the gateway side without re-flashing firmware per Unit.

The four tools deliberately stay byte-level (`scan` / `read` / `write` / `write_read`) instead of register-level wrappers. "Register" is a chip-specific concept, not an I2C protocol primitive; chip-specific drivers belong on the gateway / host side.

## Design note: a separate I2C controller for Port A

The stackchan board's existing `i2c_bus_` (= `AUDIO_CODEC_I2C_*`, GPIO 12/11, I2C controller 1) is the **internal-device shared bus** — PMU, IO ext., touch panel, IMU, audio codec all sit on it. Exposing this bus through user-callable tools risks bricking the PMIC (an errant write to AXP2101 can put the chip in an unrecoverable state).

This PR uses **I2C controller 0** on the Port A pins (GPIO 2/1) as an independent master bus. On-board ICs are unreachable from the new tools **by construction** — physical bus separation, no runtime protection list needed.

Side note: no other board in the xiaozhi-esp32 codebase currently initializes a Port A I2C bus (verified via grep across `firmware/main/boards/`). I'm only touching the stackchan board here; other board maintainers can apply the same pattern if their use cases call for it.

## Changes

**`firmware/main/boards/stackchan/config.h`**:
- Add `PORT_A_I2C_SDA_PIN` (`GPIO_NUM_2`) and `PORT_A_I2C_SCL_PIN` (`GPIO_NUM_1`)
- Comment block on `AUDIO_CODEC_I2C_*` clarifying those pins are the internal-device shared bus

**`firmware/main/boards/stackchan/stackchan.cc`**:
- Add `port_a_i2c_bus_` member to `StackChanBoard` alongside the existing `i2c_bus_`
- Add `InitializePortAI2c()` (I2C controller 0, `glitch_ignore_cnt = 7`, internal pull-ups as a fall-back; Grove Units typically carry their own 10 kΩ pull-ups)
- Call `InitializePortAI2c()` from the constructor right after `InitializeI2c()`
- Add 4 tools in `RegisterMcpTools()`:
  - `self.i2c.scan` — probe 0x08..0x77 (reserved ranges excluded) with 200 ms per-probe timeout
  - `self.i2c.read(addr, n_bytes)` — for protocols that latch output without a preceding write
  - `self.i2c.write(addr, bytes)` — `bytes` uses the Array type from the dependency PR
  - `self.i2c.write_read(addr, write_bytes, n_bytes)` — write + Repeated Start + read in a single transaction
- `scl_speed_hz = 400 kHz` — compatible with common M5Stack Units
- Transient `i2c_master_dev_handle_t` per call (`add_device` → transaction → `rm_device`)

No changes to other boards.

## Verification

Tested with an M5Stack ENV III Unit (SHT30 0x44 + QMP6988 0x70) connected to Port A on a Core S3.

**Scan** — external Unit visible, internal ICs invisible:

    $ self.i2c.scan
    { "ok": true, "addresses": [68, 112] }   # 0x44 = SHT30, 0x70 = QMP6988

**QMP6988 chip ID** (write_read pattern, register 0xD1):

    $ self.i2c.write_read addr=0x70 write_bytes=[0xD1] n_bytes=1
    { "ok": true, "bytes": [92] }            # 0x5C = QMP6988 chip ID ✓

**SHT30 high-repeatability measurement** (write + wait + read pattern):

    $ self.i2c.write addr=0x44 bytes=[0x2C, 0x06]
    { "ok": true }
    (15 ms wait)
    $ self.i2c.read addr=0x44 n_bytes=6
    { "ok": true, "bytes": [0x70, 0xC6, 0x5B, 0x77, 0x3B, 0x3B] }
    # Sensirion §4.13 compensation: T = 32.07 °C, H = 46.58 %

**QMP6988 forced-mode pressure**:

    $ self.i2c.write addr=0x70 bytes=[0xF4, 0x31]   # forced mode + osr_t=1x, osr_p=8x
    { "ok": true }
    (50 ms wait)
    $ self.i2c.write_read addr=0x70 write_bytes=[0xF7] n_bytes=6
    # Applying datasheet compensation pipeline: 1010.4 hPa ✓

Internal ICs on the shared bus (0x21, 0x23, 0x34, 0x38, 0x40, 0x41, 0x50, 0x51, 0x58, 0x68, 0x69, 0x6F) remain invisible to `self.i2c.scan` — physical bus separation behaves as designed.

`idf.py build` clean: `xiaozhi.bin` 0x30e1b0 bytes, 22% free, no warnings introduced.

## Limitations

- Only the stackchan board is covered. Other boards (`m5stack-core-s3`, `atom-*`, etc.) need the same Port A init pattern applied separately. A future refactor could extract a shared helper if multiple boards adopt this
- UART and GPIO equivalent generic tools are not in this PR. They're a natural follow-up using the same Array property foundation
- Very large `n_bytes` reads close to 256-byte upper bound are supported (25-byte calibration block reads verified) but have not been exhaustively timed across the full range

## Test plan

- [x] `idf.py build` clean (no new warnings)
- [x] All 4 tools verified against ENV III (SHT30 + QMP6988) on Port A
- [x] Internal ICs unreachable from the new tools
- [x] Existing stackchan tools (move_head / set_avatar / set_blink / take_photo / etc.) unaffected
- [ ] Verify on additional boards (out of scope; left to per-board maintainers)

### DIFF
--- a/.github/workflows/firmware-release.yml
+++ b/.github/workflows/firmware-release.yml
@@ -37,6 +37,76 @@ jobs:
           persist-credentials: false
           submodules: recursive
 
+      - name: Verify CHANGELOG.md has a dated section for the firmware tag
+        # Re-publish gate that closes the documentation debt that the
+        # CHANGELOG `[Unreleased]` Firmware section accumulated under
+        # firmware-v1.4.1 / v1.5.0 / v1.6.0 / v1.7.0 — each of those
+        # releases shipped binaries to GitHub Releases without ever
+        # promoting the matching `[Unreleased]` entries into a dated
+        # CHANGELOG section, so the in-repo CHANGELOG diverged from the
+        # artifact-of-record. From firmware-v1.8.0 onward, this step
+        # refuses to build the release if CHANGELOG.md does not yet
+        # contain a `## [firmware-vX.Y.Z] - YYYY-MM-DD` line matching
+        # the tag being pushed. Promote the entries belonging to the new
+        # tag in a prep PR, merge it, then push the tag at the merged
+        # commit. See `CONTRIBUTING.md` `Releasing the firmware`.
+        #
+        # workflow_dispatch dry-runs skip this gate because there is no
+        # tag to validate; those runs are intended for smoke-testing the
+        # build pipeline from any branch.
+        if: github.event_name == 'push'
+        run: |
+          set -euo pipefail
+
+          if [[ "$GITHUB_REF" != refs/tags/firmware-v* ]]; then
+            echo "::error::Firmware release tags must start with 'firmware-v' (got: $GITHUB_REF)"
+            exit 1
+          fi
+          TAG="${GITHUB_REF#refs/tags/}"
+
+          # Escape regex metacharacters in TAG so the version dots in
+          # tags like `firmware-v1.8.0` are matched literally rather
+          # than as ERE wildcards. Without this, a typo section header
+          # like `## [firmware-v1x8x0] - 2026-05-20` would falsely
+          # satisfy the gate for the tag `firmware-v1.8.0` (any single
+          # character at the dot positions would match).
+          TAG_ESC="${TAG//./\\.}"
+
+          # Count exact-match dated sections. Requires line layout
+          # `## [<TAG>] - YYYY-MM-DD` with no leading or trailing
+          # content. Uses `-c` so we can also fail on duplicates;
+          # `-q` would only check existence and would let a file with
+          # two matching sections pass, which leaves release-notes
+          # extraction ambiguous.
+          COUNT=$(grep -cE "^## \[${TAG_ESC}\] - [0-9]{4}-[0-9]{2}-[0-9]{2}$" CHANGELOG.md || true)
+
+          if [ "$COUNT" -eq 0 ]; then
+            echo "::error::CHANGELOG.md is missing a dated section for ${TAG}."
+            echo ""
+            echo "Expected to find exactly one line matching:"
+            echo "    ## [${TAG}] - YYYY-MM-DD"
+            echo ""
+            echo "Promote the entries that belong to ${TAG} from [Unreleased] into a"
+            echo "dated section, update the comparison-links block at the bottom of"
+            echo "CHANGELOG.md, open a PR, merge it, and re-push the tag at the merged"
+            echo "commit. See CONTRIBUTING.md 'Releasing the firmware'."
+            exit 1
+          fi
+
+          if [ "$COUNT" -gt 1 ]; then
+            echo "::error::CHANGELOG.md has ${COUNT} dated sections for ${TAG}, expected exactly 1."
+            echo ""
+            echo "Multiple dated sections for the same tag make the release-notes"
+            echo "extraction step (sed -n '/^## \[${TAG}\]/,/^## \[/p') ambiguous and"
+            echo "indicate either a copy-paste mistake during the release-prep PR or a"
+            echo "merge conflict that was resolved by keeping both sides. Collapse the"
+            echo "duplicates into a single section, open a PR, merge it, and re-push"
+            echo "the tag at the merged commit."
+            exit 1
+          fi
+
+          echo "CHANGELOG entry for ${TAG} verified (exactly 1 dated section)."
+
       - name: Build firmware with ESP-IDF Docker image
         # Same build command the regular CI Build workflow uses, so a
         # tag-driven release artifact is byte-identical to what the PR
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -98,6 +98,17 @@ jobs:
           print("Version policy: OK (pre-release allowed; local version rejected)")
           PY
 
+      - name: Verify gateway/uv.lock is in sync with gateway/pyproject.toml
+        # Catches the failure mode that motivated this step: a release
+        # bump to gateway/pyproject.toml without re-running `uv lock`,
+        # leaving uv.lock pinned at the previous version. `uv sync
+        # --frozen` reads the lock as-is and does not detect this
+        # drift; `uv lock --check` re-resolves and exits non-zero on
+        # any mismatch, blocking the publish job before it produces a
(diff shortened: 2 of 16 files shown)


kisaragi-mochi (commented):
Thanks again for the careful implementation and the prior self-codex review. The Port A bus separation design (controller 0 vs controller 1) is a clean structural way to guarantee internal-IC isolation, the verification log with the ENV III Unit (SHT30 + QMP6988) is concrete and falsifiable down to the chip-ID byte, and the byte-level tool API choice over chip-specific register wrappers is the right boundary.

A local adversarial review surfaced three items — one we'd like to ask you to fold in before landing, two as `Suggested` improvements. Conventional Comments framing below.

### Recommended for landing

**(1) Reserved I2C address ranges should be enforced in `read` / `write` / `write_read` as well** (`firmware/main/boards/stackchan/stackchan.cc:5171-5174` + gateway schema)

The `scan` contract correctly limits discovery to 0x08–0x77 (excluding reserved 7-bit ranges), but `read` / `write` / `write_read` accept `addr` 0–0x7F. That leaves the General-call address (0x00) and other reserved ranges reachable through generic writes. Even on the external Port A bus, a General-call write affects every attached device on the bus simultaneously — an accidental `write(addr=0, ...)` from an LLM or human user could flip state across multiple connected Units at once.

Suggested fix: constrain `addr` to 0x08–0x77 in both the firmware tool property range and the gateway schema, and add handler-side validation so direct callers can't bypass it. This keeps the scan and direct-call contracts consistent.

The fix should be ~1–2 lines per tool. If easier on your side, we can also land this as a small follow-up commit — your preference.

### Suggested (optional improvements)

**(2) `i2c.scan` can hold the main MCP task for up to ~22 seconds** (`firmware/main/boards/stackchan/stackchan.cc:5146-5156`)

Sequential probe of 0x08–0x77 (112 addresses) at 200 ms per probe can total ~22 s in worst-case bus-stuck scenarios. The gateway's 10 s timeout returns to the client first, while the firmware stays busy completing the remaining probes. During that window, other MCP tools (avatar, head movement, LEDs) wait. For a stackchan operated by a single user this is acceptable (scan is an infrequent discovery tool), but multi-client / SAIverse-style parallel use would feel the lag.

Mitigations: smaller per-address timeout + total scan deadline, return partial results when the bus appears stuck, or move scan off the main task. Not blocking ship — file under future polish.

**(3) Write payload arrays are unbounded** (`firmware/main/boards/stackchan/stackchan.cc:5223-5251` — same root as PR #195 Suggested item 2)

`bytes` / `write_bytes` have no `maxItems` cap before allocation and `i2c_master_transmit`. Same defense-in-depth note as the PR #195 review. A 256-item cap (matching the existing `n_bytes` read cap) would round out the symmetric contract.

---

If the addr-range fix is easy to land on your side, that would close the consistency gap nicely. The other two are at your discretion — skip them or fold them in as fits your schedule.

We'll plan to merge PR #195 first (as the foundation dependency), then this PR; happy to coordinate the rebase step if needed once #195 lands — typical flow is `gh pr edit 196 --base main` → `git rebase main` → `git push --force-with-lease`. Let us know if you'd prefer we handle the rebase.

Thanks!

maha0525:
Thanks! Addressed in commit `c808fa2` (which sits on top of a merge from the updated PR #195 branch picking up `max_items` support):

- **Recommended item 1 (addr range)**: `addr` is now restricted to 0x08..0x77 across `self.i2c.read` / `write` / `write_read` for contract symmetry with `self.i2c.scan`. Tool descriptions updated so callers see the constraint in the LLM tool surface.
- **Suggested item 3 (write payload cap)**: `bytes` / `write_bytes` capped at `max_items = 256`, matching the existing `n_bytes` read cap. Uses the new `Property::set_max_items` from PR #195.
- **Suggested item 2 (scan main-task hold)** intentionally deferred — the mitigations (per-address shorter timeout + total scan deadline / partial results on stuck bus / off-task probe) are design changes worth their own issue; landing correctness fixes first and we can iterate on scan polish separately.

Re: rebase coordination after #195 lands — happy to handle the rebase locally with `git rebase main && git push --force-with-lease` once you've merged #195, or you can use GitHub's "Update branch" again. Either works.

Build verified clean.

maha0525:
For visibility: filed item 2 (scan main-task hold) as #201, with the three suggested mitigation paths captured. Design-change scope; landing the correctness fixes first.

kisaragi-mochi:
Merged via 3a3569f. Thanks for the address-range enforcement across `self.i2c.read` / `write` / `write_read`, the `max_items = 256` cap on `bytes` / `write_bytes`, and the carve-out of item 2 as #201 with the three mitigation paths captured.

Treating Port A as I2C controller 0 (independent from the internal controller-1 bus) is a clean structural way to isolate the on-board safety-critical ICs from the externally exposed bus — the contract becomes inherent to the hardware boundary rather than enforced only by software discipline.

A small follow-up landed as PR #202 to align the gateway-side JSON Schema `addr` range (`minimum: 8, maximum: 119`) with the firmware-side property range. That was carved out of this review and picked up here directly to keep your turnaround clean rather than asking for another revision round.

Thanks again for the thorough work, and looking forward to more.

PR #205, opened 2026-05-21
feat(firmware/ws): fail-fast on invalid server hello (#191)

kisaragi-mochi:
## Summary

Fixes the ~10 s per-candidate reconnect delay surfaced in Codex adversarial review of #187 (Round 5). When a server hello is malformed (missing/non-string `transport`, missing/empty `session_id`, or unsupported `transport`), the firmware now falls back to the next candidate URL within ~100 ms instead of waiting out the full 10 s `WEBSOCKET_PROTOCOL_SERVER_HELLO_EVENT` timeout.

Closes #191.

## Approach

- New event bit `WEBSOCKET_PROTOCOL_SERVER_HELLO_FAILED (1 << 1)` in `websocket_protocol.h`, adjacent to the existing `_EVENT (1 << 0)`.
- `ParseServerHello()` sets `_FAILED` on each of its 3 early-return paths before logging and returning. The existing `ESP_LOGE` lines are kept for diagnostics.
- `OpenAudioChannelInternal()`'s `xEventGroupWaitBits` now waits on `(_EVENT | _FAILED)` with the same 10 s timeout, and dispatches one of three outcomes:
  1. `_EVENT` set → continue with the existing happy path (byte-for-byte unchanged).
  2. `_FAILED` set → `ESP_LOGW` "Server hello rejected by candidate i/N; falling back to next candidate", disarm `notify_disconnect` (mirror of the intentional-teardown pattern in `CloseAudioChannel` / destructor / prologue), `websocket_.reset()`, and `continue` to the next candidate URL.
  3. Neither bit set within 10 s (no hello at all) → retain the existing timeout-handling path unchanged.
- `xEventGroupClearBits` calls at the prologue and at the top of the candidate loop are updated to clear both bits so no stale state leaks between candidates.

## Outcomes mapped to acceptance criteria

| Criterion | Outcome |
|---|---|
| Malformed hello → fallback within ~100 ms (not 10 s) | ✅ `_FAILED` branch returns to candidate loop immediately |
| Happy path (valid hello → audio channel opens) unchanged | ✅ `_EVENT` branch is unchanged; `ParseServerHello` success path still hits `SetBits(_EVENT)` |
| No-hello timeout path (10 s) unchanged | ✅ `if (!(bits & _EVENT))` branch retains the existing log + reset + continue |
| License boundary preserved | ✅ Edits limited to `firmware/main/protocols/`; `firmware/main/boards/stackchan/` (GPL-3.0) untouched |

[...]

PR #206, opened 2026-05-21
fix(firmware/stackchan): readable touch event log (format bug + per-channel raw decode)

maha0525:
## Summary

The TAP / STROKE log line in the Si12T head-touch driver currently emits
`duration=lums raw=0xNNNN` instead of human-readable values. This PR
fixes the log so it actually tells you what the touch was.

## Why it matters

Without this fix, both fields are effectively garbage:

```
touch event: STROKE (zones=000 duration=lums raw=0x1130)
touch event: STROKE (zones=000 duration=lums raw=0x1F3)
```

I assumed for a while that the trailing `raw=0x1130` (4400) was an actual
sensor reading and that the false strokes I was chasing had a raw-value
signature distinguishing them from real touches. They don't — the value
shown there is `duration_ms`'s low byte, not the Si12T register.

## Root cause

ESP-IDF's nano-printf does not parse `%llums`. The literal `lums` prints
verbatim and the va_arg pointer never advances over `duration_ms`, so the
following `%02X` consumes the long-long's bytes instead of
`last_output1_raw_`. `zones=000` is also misleading: the snapshot is
overwritten every poll tick, so by the time `HandleTap` / `HandleStroke`
fires on the falling edge those fields always read the post-release zero
state, not what the user actually touched.

## What this PR does

- Switch `%llums` to `%u ms` after casting to `unsigned`. Matches the
  `%ums` style already used elsewhere in this file (e.g. motion driver
  duration logs) and side-steps the `ll` modifier limitation.
- Capture the rising-edge sensor state into `press_start_zones_` /
  `press_start_output1_raw_` so the log can report what the touch
  actually looked like, not what it looked like after the user lifted.
- Extend the log line with `start_zones=NNN start_raw=0xNN ch=CCCC
  release_raw=0xNN`. The `ch=CCCC` field decodes Output1's four 2-bit
  channel levels as `0/L/M/H`. CH4 is unwired on stack-chan, so a non-`0`
  CH4 character is an immediate flag for wiring noise / EMI.
- Thread `duration_ms` through `HandleTap` too, so 400-600 ms grey-zone
  presses are visible in the log instead of being silently logged as TAP
  with no timing.

## Before / after

```
# Before
touch event: STROKE (zones=000 duration=lums raw=0x1F3)

# After (deliberate single-zone stroke on CH1)
touch event: STROKE start_zones=100 start_raw=0x03 ch=H000 release_raw=0x00 duration=1200 ms
```

## Test plan

- [x] Builds clean with ESP-IDF v5.5.4 (`idf.py build`, no warnings,
      same binary size class as before).
- [x] Flashed to hardware via `idf.py flash`. Deliberate TAP and STROKE
      events on CH1 now produce the new readable format with correct
      duration values, raw byte that matches the touched zone, and
      `release_raw=0x00` at the falling edge.
- [ ] Reviewer hardware test (optional): trigger TAP and STROKE,
      confirm the log line is parseable.

## Notes

This is purely a logging change — no touch detection logic or timing
constants were modified. The new fields are additive; if a downstream
tool greps for `touch event: STROKE` followed by `duration=` it will
still match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,29 @@ documented-only.
 
 ### Firmware
 
+- Fixed: the Si12T head-touch driver's TAP / STROKE log line printed
+  `duration=lums raw=0xNNNN` instead of human-readable values, because
+  ESP-IDF's nano-printf cannot parse the `%llums` specifier and the
+  failed parse stops `va_arg` from advancing past `duration_ms` — so
+  the following `%02X` consumed the long-long's bytes rather than the
+  Si12T Output1 register. `zones=000 raw=0x00` was also misleading
+  because the snapshot fields were overwritten every poll tick and
+  read the post-release zero state by the time the falling-edge
+  handler logged. The log line now uses `%u ms`, captures the
+  rising-edge sensor state separately into `press_start_*`, and reports
+  it as
+  `start_zones=NNN start_raw=0xNN ch=CCCC release_raw=0xNN duration=NN ms`,
+  with `ch=CCCC` decoding Output1's four 2-bit channel levels as
+  `0/L/M/H` (CH4 should always be `0`; a non-`0` CH4 character flags
+  wiring noise / EMI). `HandleTap` also receives the duration so the
+  400-600 ms grey-zone TAPs print their timing. The rising-edge
+  capture also fires on the cooldown-suppressed branch so a touch
+  whose press started during the post-reaction cooldown but released
+  after cooldown expired logs its own start state instead of the
+  previous touch's. Purely a logging change — touch detection logic
+  and timing constants are unchanged. Contributed via
+  [PR #206](https://github.com/kisaragi-mochi/stackchan-mcp/pull/206).
+
 - Fixed: a server-side close arriving between the WebSocket server
   hello and the main task resuming no longer cancels the reconnect
   timer that the per-socket `OnDisconnected` lambda just armed.
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -624,6 +624,15 @@ private:
     uint64_t   last_event_us_ = 0;
     bool       last_zone_snapshot_[3] = {false, false, false};
     uint8_t    last_output1_raw_ = 0;
+    // Press-start snapshot. last_* fields above are overwritten every poll
+    // tick, so by the time HandleTap / HandleStroke fires on the falling edge
+    // they reflect the release state (zones=000 raw=0x00). press_start_*
+    // captures the rising-edge state so the log can show what the sensor
+    // actually saw when the touch began. Useful for distinguishing genuine
+    // touches (CH1〜CH3 set) from false positives (e.g. CH4 noise, raw=0x00
+    // with press judged via debounce, etc.).
+    bool       press_start_zones_[3] = {false, false, false};
+    uint8_t    press_start_output1_raw_ = 0;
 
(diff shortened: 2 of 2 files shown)


kisaragi-mochi (commented):
## Review summary

Solid bug fix with a clean root-cause writeup. The `%llums` /
nano-printf interaction is correctly identified, and the press-start
snapshot extraction is a nice side-fix for the misleading
`zones=000 raw=0x00` artifact (the snapshot was reading the
post-release state by the time the falling-edge handler logged).

I ran a focused review and have two requests before merge:

### 1. Rebase + restore the existing `[Unreleased]` entries

The PR currently shows `CONFLICTING` because `CHANGELOG.md` removes
the existing `[Unreleased] Firmware` and `[Unreleased] Gateway`
entries that landed since this branch was cut
(#186 / #195 / #196 / #202 / #205). If the next release is promoted
from the current `[Unreleased]`, those already-landed changes would
drop out of the release notes.

Could you rebase onto current `main` and add the new firmware log
entry on top of the existing bullets? `stackchan.cc` should not
conflict.

### 2. Also capture `press_start_*` in the cooldown-rising-edge branch

In `PollTouch()` at the rising-edge cooldown branch
(`stackchan.cc:3547-3552`), `touch_pressed_prev_` and
`touch_press_start_us_` are updated but `press_start_*` is not.
If a new touch begins while still in the post-reaction cooldown
and is held until cooldown expires, the falling-edge handler will
emit TAP/STROKE with the *previous* touch's `start_zones` /
`start_raw` — exactly in the repeated-touch / noise overlap
scenario this logging is meant to clarify.

A small additional capture in that branch resolves it:

```cpp
if (now_us < cooldown_until_us_) {
    touch_pressed_prev_ = now;
    touch_press_start_us_ = now_us;
    press_start_zones_[0] = s.zone[0];
    press_start_zones_[1] = s.zone[1];
    press_start_zones_[2] = s.zone[2];
    press_start_output1_raw_ = s.output1_raw;
    return;
}
```

### Other notes

- `LogTouchEvent` extraction + `ch=CCCC` decode makes both call
  sites identical and adds useful debug info. CH4-as-EMI-canary on
  stack-chan boards (the head has 3 zones) is a useful interpretive
  hint to bake into the log.
- `(unsigned)duration_ms` cast (nit, defer OK): a C-style cast
  matches the surrounding ESP-IDF code; `static_cast<unsigned>`
  would document the narrowing intent more explicitly. Either is
  fine — STROKE cooldown bounds `duration_ms` well below
  `UINT32_MAX` ms (~49.7 days).
- No regression risk in the existing logic: touch detection logic
  and timing constants are unchanged, matching the PR description.

Thanks for the careful diagnosis and the clean fix.

kisaragi-mochi (approved):
LGTM. Verified the revisions against the previous review:

1. CHANGELOG rebase: the previously deleted `[Unreleased] Firmware` entries from #186 / #195 / #196 / #202 / #205 are back in place, with the PR #206 log-fix entry added on top. `CHANGELOG check` CI is green.

2. `press_start_*` cooldown-suppressed-branch capture: the rising-edge capture in `firmware/main/boards/stackchan/stackchan.cc` now updates `press_start_zones_` / `press_start_output1_raw_` in the `now_us < cooldown_until_us_` branch as suggested, so a touch whose press starts during post-reaction cooldown but releases after cooldown expires logs its own start state instead of the previous touch's.

A second-pass codex review on the revision returned no further findings — the change stays scoped to logging and the underlying touch detection logic is unchanged.

Thanks for the careful diagnosis and the clean rebase.

PR #207, opened 2026-05-21
fix(firmware): trigger play_popup_on_listening_ from StartListening()

maha0525:
## Summary

`Application::StartListening()` is a public API called from board-level touch
buttons, gateway-driven `listen.start`, and any other path that wants to bring
the device into listening mode. But the `OGG_POPUP` "listening started" audio
cue currently only fires on the wake-word activation paths
(`HandleWakeWordDetectedEvent` / `ContinueWakeWordInvoke`) — every other
activation source set `MAIN_EVENT_START_LISTENING` and reached the
`kDeviceStateListening` branch in `HandleStateChangedEvent`, but the
post-`ResetDecoder` `PlaySound` site there is gated on
`play_popup_on_listening_` and stayed silent.

## Fix (revised after review)

Move `play_popup_on_listening_ = true` into `HandleStartListeningEvent`,
**not** into the public `StartListening()` setter — per review feedback the
public setter should stay a thin event setter
(`xEventGroupSetBits` + return) so every write to the flag stays on the
same main task that reads and clears it in `HandleStateChangedEvent`,
matching the wake-word path that already flips the flag from the main
task before calling `ContinueWakeWordInvoke`.

The flag set sits after the `kDeviceStateActivating` /
`kDeviceStateWifiConfiguring` / null-protocol early returns and is gated
on `state == kDeviceStateIdle || state == kDeviceStateSpeaking` (= the
states that actually transition toward Listening). A no-op
`StartListening()` therefore does not latch the cue for some unrelated
future listening transition.

## Why this matters for downstream

For boards that expose touch / button activation paths (the stack-chan board
in this repo, but the pattern is generic), the silent listen activation made
the device feel unresponsive — there was no audible feedback that the user
input had been registered. The wake-word UX has had this cue from the start;
generalizing it to all activation paths is just removing an accidental
asymmetry.

## Test plan

- [x] Builds clean with ESP-IDF v5.5.2 (Docker image used by CI).
- [x] Flashed to hardware (stack-chan board, FT6336 touch panel). Touch-driven
      listen activation now plays the popup sound on listening entry,
      matching wake-word activation.
- [x] Wake-word activation still plays the cue (no double-play, the flag is
      idempotent).

## Notes

This is the first of a two-PR set for stack-chan touch-driven listen UX.
**PR-M is now open as #208** (`feat(firmware/stackchan): touch-driven listen
UX (single-shot push-to-talk)`) and handles the stack-chan board-side
changes (touch routing to `StopListening` for listening-state taps, debounce,
RGB LED feedback, etc.). PR-L is the more broadly useful half because it
benefits any board that uses touch / button activation, not just stack-chan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,25 @@ documented-only.
   close) is unchanged. Closes
   [#189](https://github.com/kisaragi-mochi/stackchan-mcp/issues/189).
 
+- Fixed: the `OGG_POPUP` listening cue was only triggered on wake-word
+  activation paths (`HandleWakeWordDetectedEvent` /
+  `ContinueWakeWordInvoke`), so callers of the public
+  `Application::StartListening()` API — board-level touch buttons,
+  server-driven listen, etc. — silently lost the audible "listening
+  started" feedback. `HandleStartListeningEvent` now arms
+  `play_popup_on_listening_` itself, after the early-return branches
+  for `kDeviceStateActivating` / `kDeviceStateWifiConfiguring` /
+  null-protocol but before the Idle / Speaking-abort listening
+  dispatches, so the flag is only latched when the function is
+  actually going to transition the device toward listening. The
+  public `Application::StartListening()` stays a thin event setter
+  (`xEventGroupSetBits` + return) so the write happens entirely on
+  the main task, matching the wake-word path that already flips the
+  flag from the main task before calling `ContinueWakeWordInvoke`.
+  No behaviour change for the wake-word path (the flag was already
+  true there before this function ran). Contributed via
+  [PR #207](https://github.com/kisaragi-mochi/stackchan-mcp/pull/207).
+
 - Fixed: WebSocket candidate fallback is now fail-fast when a server
   hello is malformed (missing/non-string `transport`, missing/empty
   `session_id`, or unsupported `transport`). A new
--- a/firmware/main/application.cc
+++ b/firmware/main/application.cc
@@ -703,6 +703,10 @@ void Application::ToggleChatState() {
 }
 
 void Application::StartListening() {
+    // Thin event setter. The popup-on-listening flag is armed inside
+    // HandleStartListeningEvent (main task) so all writes to
+    // play_popup_on_listening_ converge to the same task that reads
+    // and clears it in HandleStateChangedEvent.
     xEventGroupSetBits(event_group_, MAIN_EVENT_START_LISTENING);
 }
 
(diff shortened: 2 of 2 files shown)


kisaragi-mochi (commented):
## Review summary

Clean, well-motivated extraction of the "popup on listen" gating
flag — the wake-word path having an audible cue while every other
activation source silently entered listening mode is exactly the
kind of accidental asymmetry that's worth removing.

I ran a focused review and have two findings to discuss before merge:

### 1. Race-vs-event-contract on `play_popup_on_listening_`

`Application::StartListening()` reads as a thin event setter (post
`MAIN_EVENT_START_LISTENING` and return). Setting
`play_popup_on_listening_` directly in this function means the write
happens from whatever task called `StartListening()` (board touch
callback, WebSocket dispatch, etc.), while
`HandleStateChangedEvent()` reads and clears the same flag from the
main task. `play_popup_on_listening_` in `application.h:142` is a
plain `bool`, not `std::atomic<bool>` as the PR description
suggests. In practice the word-aligned write is unlikely to tear on
Xtensa, and the `xEventGroupSetBits` immediately after carries an
implicit FreeRTOS barrier, so the race is mostly theoretical — but
keeping `StartListening()` as a thin event setter is the cleaner
contract.

Suggested shape:

```cpp
void Application::StartListening() {
    xEventGroupSetBits(event_group_, MAIN_EVENT_START_LISTENING);
}

void Application::HandleStartListeningEvent() {
    play_popup_on_listening_ = true;  // main task only
    // ... existing body
}
```

This also matches how the wake-word paths flow into the main task
before flipping the flag, so all writes converge to one task.

### 2. Coverage and PR-L / PR-M (#208) dependency

`StartListening()` callers in `firmware/main/boards/` are all
button/touch handlers that call it directly, but the stack-chan
board doesn't appear in that list — its touch handler currently
reaches listening mode via `ToggleChatState()` → `SetListeningMode()`
without ever passing through `StartListening()`. So #207 alone
wouldn't deliver the popup cue on stack-chan touch, which is the
motivating scenario in the PR body.

Reading PR #208 (PR-M), I see this is precisely what it addresses:
it swaps `app.ToggleChatState()` → `app.StartListening()` on the
stack-chan idle/speaking touch path (with explanatory comments
about why `HandleStartListeningEvent` is the desired entry). So the
two PRs form a stack-chan-complete pair as designed; PR #207 on its
own benefits other boards that already call `StartListening()`
directly.

Two small asks:

- Mention #208 explicitly in the PR-L body's `Notes` section
  ("PR-M is now open as #208") so reviewers don't have to cross-
  reference to find the companion.
- The body still references `PR-M (separate, not yet opened)`; this
  was true at #207 open time but #208 opened ~8 minutes later, so
  a short edit avoids future confusion.

### Other notes

- Builds clean and the wake-word path is unaffected (its existing
  set runs before the event dispatches anyway).
- `play_popup_on_listening_` is auto-cleared in
  `HandleStateChangedEvent` (`application.cc:950-951`) so the flag
  doesn't leak across activations.
- `CHANGELOG.md` shows a `CONFLICTING` state because the entry is
  positioned before the existing `[Unreleased] Firmware` entries
  that landed since this branch was cut (#186 / #195 / #196 / #202
  / #205). A simple rebase on current `main` should resolve it.
  (Also the `[PR #TBD-L]` placeholder in the entry probably wants
  to become `[PR #207]` on the rebase.)

Thanks for the clear motivation and writeup.

kisaragi-mochi (approved):
Re-reviewed the revision against the previous review feedback — both findings from the earlier round are addressed cleanly: `Application::StartListening()` is back to a thin event setter, `play_popup_on_listening_` is now set inside `HandleStartListeningEvent()` (main task only) after the early-return branches, and PR #208 is referenced from the Notes section.

A second-pass codex review on the updated branch surfaced one additional finding (popup flag state-leak when `OpenAudioChannel()` fails or the Listening transition is aborted), which is intentionally carved out as a separate follow-up issue (#220) rather than gating this PR.

The reasoning for the carve-out: the underlying state-machine question — how `play_popup_on_listening_` should behave when the Listening transition does not complete — is broader than the popup-cue API extension that this PR was reviewed for. The current change does not introduce that leak window (an equivalent window existed orthogonally on the wake-word path before this PR; the new path inherits the same shape). The focused fix belongs in the follow-up, not in this PR's diff.

Approving on that basis.

Note: this branch is currently behind `main` — #219 / #205 etc. have landed since it was cut, and `strict` branch-protection requires the branch to be up to date before merge. "Update branch" on the GitHub UI, or a local rebase + force-push, will unblock the status check.

Thanks for the clean revision and the thoughtful response to the contract-shape feedback.

PR #208, opened 2026-05-21
feat(firmware/stackchan): touch-driven listen UX (single-shot push-to-talk)

maha0525:
## Summary

Retargets stack-chan's FT6336 touch handling for a single-shot push-to-talk
UX (touch → listen → touch → submit) instead of xiaozhi's default
continuous-conversation model. **Stack-chan board only** — `application.cc`
and other boards are untouched.

## Motivation

stack-chan exposes a touch LCD as its primary input. The default behavior
ended up feeling broken in two ways:

1. **Listening-state tap aborts capture instead of submitting**.
   `ToggleChatState()` → `CloseAudioChannel()` tears down the WebSocket,
   which makes the gateway treat the in-flight capture as `aborted_mid_capture`
   and discard it. A user trying to "tap to submit" loses their utterance.

2. **TTS finish silently re-arms listening**. With `AutoStop` (= upstream
   default via `ToggleChatState`), the device transitions back to Listening
   right after `tts.stop`. The user's next "start" tap therefore lands on
   `listen.stop` and the model is broken.

Both behaviors make sense for xiaozhi's continuous voice assistant model.
For a touch-driven device, they make the UX feel non-deterministic.

## Changes

All in `firmware/main/boards/stackchan/stackchan.cc`:

- Listening-state short tap → `StopListening()` (= `listen.stop`,
  record-then-flush) instead of `CloseAudioChannel()` (= WS teardown).
- Idle / speaking short tap → `StartListening()` (forces `ManualStop`)
  instead of `ToggleChatState()` (uses `AutoStop`).
- 300 ms post-release debounce on FT6336 chatter.
- 30 s auto-`StopListening()` if listening is entered and never released
  (user walked away).
- RGB LED tap feedback (dim green on activation, off on submit) via a
  board-local `SetAllRgbLeds` helper that shares the I2C path used by the
  `set_all_leds` MCP tool. MCP can still overwrite.
- `%lld` + concatenated literal → `(int)cast` + `%d` on the duration log,
  matching the nano-printf-safe pattern already used by this file's motion
  driver.

## Related

Pairs with PR #207 (StartListening popup cue) for full audible+visual
activation feedback. Does **not** depend on it — this PR is functional alone,
just without the audible cue on touch activation.

## Test plan

- [x] Builds clean with ESP-IDF v5.5.2 (Docker image used by CI).
- [x] Flashed to hardware (stack-chan board). Touch activates listening with
      green LED feedback, second touch submits the captured audio with LED
      off, debounce suppresses double-press chatter, idle 30 s auto-stops
      listening.
- [x] Other-board behavior is unchanged (`application.cc` not touched).

## Notes

The strategy doc for this fork lists this work as PR-M; the upstream-facing
description is consolidated here as a single commit for review. Internal
debug `ESP_LOGI` lines that were used while developing this feature are
stripped from the PR (kept locally on dev/integration for future
reproduction debugging).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,29 @@ documented-only.
   true there before this function ran). Contributed via
   [PR #207](https://github.com/kisaragi-mochi/stackchan-mcp/pull/207).
 
+- Added (stackchan): touch-driven listen UX overhaul. The FT6336
+  short-tap behavior now models stack-chan as a single-shot
+  push-to-talk device (touch → listen → touch → submit) rather than
+  xiaozhi's default continuous-conversation model. Concretely: (a) a
+  short tap while in `kDeviceStateListening` now calls
+  `StopListening()` instead of `CloseAudioChannel()`, so the gateway
+  receives `listen.stop` (= record-then-flush) instead of a transport
+  teardown that aborts the in-flight capture; (b) a short tap from
+  idle / speaking now calls `StartListening()` (which forces
+  `ManualStop`) instead of `ToggleChatState()` (which uses
+  `AutoStop`), so the device does not re-arm listening immediately
+  after `tts.stop` and the next tap is reliably interpreted as a new
+  `listen.start`; (c) 300 ms press-after-release debounce on
+  FT6336 chatter; (d) 30-second auto-`StopListening` for users who
+  start a listen and then walk away; (e) RGB LED tap feedback (green
+  on activation, off on submit) via a small `SetAllRgbLeds` helper
+  that shares the I2C path used by the `set_all_leds` MCP tool;
+  (f) `%lld` → `%d (int)cast` for the existing duration log,
+  matching the nano-printf-safe pattern already used in this file's
+  motion driver. Stack-chan-only — other board UX is unchanged.
+  Contributed via
+  [PR #208](https://github.com/kisaragi-mochi/stackchan-mcp/pull/208).
+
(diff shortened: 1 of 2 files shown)


kisaragi-mochi:
Thanks for this — merged in 44211ab.

About the force-push you'll have seen on this branch (2026-05-26): that rebase was on our end, nothing you need to act on. After this PR sat in review, we landed several other commits touching `stackchan.cc` (#196 Port A I2C, #206 touch log, #223 Port B WS2812, #210 AvatarSet), which put the branch into conflict on `CHANGELOG.md`. That was on us for not coordinating the merge order, so we rebased your branch onto the latest main on our end — your touch-UX changes auto-merged cleanly, only the CHANGELOG entry needed reordering (and the `PR #TBD-M` placeholder was filled in to `#208`). Your authorship is preserved; only the commit hashes changed.

Verified on real hardware: the touch-driven listen UX works as designed — a short tap enters listening with the green LED cue, a second tap submits via `StopListening()` so the WebSocket stays connected (no `aborted_mid_capture`), and the device returns to idle. Confirmed against the latest main (mDNS gateway discovery + the touch flow).

One item is tracked separately as #233: the 30s auto-stop also applies to listening sessions started outside the touch path (wake-word / realtime). No impact in the current touch/MCP-driven setup; noted for when wake-word config becomes locally configurable.

Thanks again for the clean single-shot push-to-talk design.

Issue #169, opened 2026-05-18
Firmware should separate persistent WebSocket MCP transport from logical audio/listening state


1nwooozip:
## Summary

The stackchan-mcp firmware currently treats the WebSocket connection as an on-demand audio channel, but the gateway also uses the same WebSocket as the MCP control channel.

This causes two related problems:

1. On boot, the device reaches idle without connecting to the configured WebSocket gateway.
2. If the firmware is locally patched to connect the WebSocket during startup and keep it alive for MCP control, the app can get stuck in `listening` because `IsAudioChannelOpened()` uses the physical WebSocket connection state as the logical audio/listening state.

The underlying issue seems to be that the firmware does not separate:

- physical WebSocket transport connection
- logical audio/listening session state

## Environment

- Device: M5Stack StackChan / CoreS3 ESP32-S3
- Firmware: stackchan-mcp prebuilt `merged-binary.bin`
- Firmware version shown in boot log: `2.2.6`
- Gateway: `stackchan-mcp` v0.7.0
- Gateway URL configured in NVS: `ws://<gateway-lan-ip>:8765/`
- Device and gateway are on the same LAN
- `CONFIG_DISABLE_OTA_WEBSOCKET_CONFIG=y`

## Observed behavior 1: no WebSocket connection attempt on boot

Wi-Fi connects successfully and OTA / activation completes:

```text
I WifiStation: Found AP: <ssid>, RSSI: <rssi>
I wifi:connected with <ssid>
I WifiStation: Got IP: <device-lan-ip>
I Application: Network connected
I StateMachine: State: starting -> activating
I Ota: Current version: 2.2.6
I Ota: OTA response includes a websocket section; ignored (CONFIG_DISABLE_OTA_WEBSOCKET_CONFIG=y).
I Ota: Current is the latest version
I Application: Activation done
I StateMachine: State: activating -> idle
```

But there is no WebSocket connection attempt at all.

Expected logs such as these never appear:

```text
Adding websocket gateway candidate from websocket.url
Connecting to websocket server
No websocket gateway URL configured
```

The gateway does not discover the device while the device is idle.

## Confirmed

- Wi-Fi is connected.
- The gateway is running and listening on port `8765`.
- The configured gateway URL is stored in NVS as `websocket.url = ws://<gateway-lan-ip>:8765/`.
- The firmware registers the StackChan MCP tools on boot.
[...]

PR #209, opened 2026-05-21
feat(gateway): forward device-driven listen captures to an external HTTP hook

maha0525:
## What

Forward device-initiated listen captures — any path that calls
`Application::ToggleChatState` / `WakeWordInvoke` / `StartListening` on the
firmware: wake word, button press, LCD touch — to an externally configured
HTTP hook as an Ogg/Opus payload. **Default OFF, opt-in via env var.**

## Why

stackchan-mcp's primary listen model today is MCP-client-driven (the
`listen()` tool): the LLM agent decides when to open the device's microphone
and the gateway transcribes the resulting Opus stream through a registered
STT engine. This works well for "AI agent initiates listening" workflows.

The device-side firmware (inherited from xiaozhi-esp32) also has a reverse
path: when a wake word fires / a button is pressed / the LCD is tapped, the
firmware sends `{\"type\":\"listen\",\"state\":\"start\"}` to the gateway and
starts streaming Opus frames. The gateway currently ignores these inbound
listen messages — frames are dropped at `audio_stream.handle_audio_frame`'s
\"no active recording slot\" branch. The existing comment documents this:

> the device may emit audio on its own (e.g. after an autonomous wake-word
> detection) and the gateway has no STT pipeline running for those frames yet.

This PR fills that gap **for operators who want it** — without changing the
MCP-driven default. Use cases: a non-Whisper recognizer, a recorder, or
(our case) a Gemini-powered persona that consumes audio as `inline_data`.

## How

1. **Inbound listen handler** (`esp32_client._handler`): when
   `STACKCHAN_AUDIO_HOOK_URL` is configured AND no MCP-driven `listen()` is
   already capturing, open the shared `audio_stream` recording slot on
   `state=\"start\"` and close it on `state=\"stop\"`. Without the hook URL,
   the device-driven branch logs at debug and returns immediately — current
   behavior preserved.

2. **Ogg/Opus packing** (`audio_input_hook.pack_opus_frames_to_ogg`): pure
   Python RFC 7845 + RFC 3533. No new runtime dependencies — the existing
   `opuslib` is for the codec, not the container, and we intentionally
   avoid pulling in `pyogg` for a 200-line wrapper.

3. **HTTP push** (`audio_input_hook.push_audio_capture`): aiohttp POST with
   `Content-Type: audio/ogg`, optional `Authorization: Bearer
   <STACKCHAN_AUDIO_HOOK_TOKEN>` (falls back to `STACKCHAN_TOKEN`), and
   `X-StackChan-Session: <gateway session id>`. Fire-and-forget; failures are
   logged at WARNING and do not propagate.

4. **Disconnect cleanup**: connection close mid-capture drops the partial
   buffer (mirrors the existing session-mismatch discard logic in
   `audio_stream.handle_audio_frame`).

## Compatibility

- **Default OFF / opt-in**: `STACKCHAN_AUDIO_HOOK_URL` unset → inbound listen
  messages are still logged at debug and discarded, matching today's behavior.
- **No conflict with MCP-driven `listen()`**: if an MCP `listen()` has
  already opened the recording slot, the device-driven branch defers (existing
  slot is honored). The shared `audio_stream` module-level singleton remains
  the single capture buffer.
- **No new runtime dependencies**.

## Test plan

- [x] `tests/test_audio_input_hook.py` (new, 9 cases): Ogg page structure
      (BOS / OpusTags / EOS, granule monotonicity, multi-page layout), CRC
      round-trip (parse, zero, recompute, compare), HTTP push success /
      empty / 5xx.
- [x] `tests/test_esp32_client.py` (extended, 3 new cases): device-driven
      listen → frame → stop → hook fire; hook URL absent → no recording
      slot; disconnect mid-capture → partial buffer discarded, no hook.
- [x] All pre-existing tests continue to pass.
- [x] End-to-end verified on hardware (stack-chan board): touch → listen →
      touch → submit → POST received downstream with a valid Ogg/Opus file
      that decodes to the captured utterance.

## Out of scope

- Audio recognition itself — this PR ships frames out the door; recognition
  is the receiver's responsibility.
- Multiple concurrent hooks / topic routing. One configured URL gets every
  device-driven capture.

## Related

- Existing comment in `audio_stream.handle_audio_frame` documenting the
  drop point (\"no active recording slot\")
- Issue #8 (Phase 4: Opus audio stream)
- PR #136 / #169 (WS lifecycle around CloseAudioChannel / device-initiated
  audio) — orthogonal, but co-exists cleanly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -242,6 +242,22 @@ documented-only.
 
 ### Gateway
 
+- Added: optional device-driven listen audio capture forwarding via
+  `STACKCHAN_AUDIO_HOOK_URL`. When set, inbound device-initiated
+  listen captures (wake-word, button, LCD touch — any path that calls
+  `Application::ToggleChatState` / `WakeWordInvoke` / `StartListening`
+  on the firmware) are buffered through the existing `audio_stream`
+  recording slot, packed into an Ogg/Opus container, and POST'd as
+  `Content-Type: audio/ogg` (with optional `Authorization: Bearer`
+  from `STACKCHAN_AUDIO_HOOK_TOKEN` / `STACKCHAN_TOKEN`) to the
+  configured URL. Default-OFF and opt-in: with the env var unset,
+  inbound device-driven listen frames are still logged at debug and
+  discarded as before. Coexists with MCP-driven `listen()` by sharing
+  the same single recording slot — if `listen()` is already
+  capturing, the device-driven branch defers. Pure-Python RFC 7845 /
+  RFC 3533 Ogg packing, no new runtime dependencies. Contributed via
+  [PR #TBD-F](https://github.com/kisaragi-mochi/stackchan-mcp/pull/TBD-F).
+
 - Fixed: mDNS now advertises the host's IPv4 addresses ordered by LAN
   reachability instead of in interface-enumeration order. On a host with
   multiple interfaces (a CGNAT `100.64.0.0/10` overlay address, several LAN
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -253,10 +253,14 @@ documented-only.
   configured URL. Default-OFF and opt-in: with the env var unset,
   inbound device-driven listen frames are still logged at debug and
   discarded as before. Coexists with MCP-driven `listen()` by sharing
-  the same single recording slot — if `listen()` is already
-  capturing, the device-driven branch defers. Pure-Python RFC 7845 /
-  RFC 3533 Ogg packing, no new runtime dependencies. Contributed via
-  [PR #TBD-F](https://github.com/kisaragi-mochi/stackchan-mcp/pull/TBD-F).
+  the same single recording slot — the two sources defer to each
+  other rather than clobber: a device-driven listen.start arriving
+  while `listen()` is capturing is ignored with a warning log, and
+  conversely an MCP `listen()` invoked while a device-driven capture
+  is buffering is declined with an explicit error. Pure-Python
+  RFC 7845 / RFC 3533 Ogg packing, no new runtime dependencies.
+  Contributed via
+  [PR #209](https://github.com/kisaragi-mochi/stackchan-mcp/pull/209).
 
 - Fixed: mDNS now advertises the host's IPv4 addresses ordered by LAN
   reachability instead of in interface-enumeration order. On a host with
--- a/gateway/stackchan_mcp/audio_input_hook.py
+++ b/gateway/stackchan_mcp/audio_input_hook.py
@@ -124,6 +124,33 @@ def _ogg_crc32(data: bytes) -> int:
     return crc
 
 
+def _packet_to_segments(packet: bytes) -> list[bytes]:
+    """Split an Ogg packet into 255-byte lacing segments (RFC 3533 §6).
+
+    Packets longer than 255 bytes are split into 255-byte runs; packets
+    whose length is exactly a multiple of 255 are terminated with a
+    zero-length segment so the parser knows the packet ended there
+    (otherwise it would expect a continuation into the next page).
+    Variable-bitrate Opus frames can exceed 255 bytes in practice, so
+    this split must happen before page assembly.
+    """
+    segments: list[bytes] = []
+    if not packet:
+        # An empty packet is itself a single zero-length segment.
+        return [b""]
+    pos = 0
+    n = len(packet)
+    while pos < n:
+        chunk = packet[pos:pos + 255]
+        segments.append(chunk)
+        pos += 255
+    if len(packet) % 255 == 0:
+        # Packet ends exactly on a 255-byte boundary — append a
+        # terminating zero-length segment per RFC 3533.
+        segments.append(b"")
+    return segments
+
+
 def _build_ogg_page(
     *,
     header_type: int,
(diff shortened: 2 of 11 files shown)


kisaragi-mochi (commented):
Thanks for this. Reading it alongside #208 (touch activation), #216 (preserving the self-hosted gateway URL against the OTA writeback), and the #213–#215 PCM-in stack, the intent is clear: stack-chan as the audio I/O surface for a self-hosted gateway / external system — touch-activated capture (#208) forwarded out here (#209), audio pushed back in (#213–#215). A deliberate, coherent integration.

One positioning question, then one change request before merge.

**Positioning:** is the external-hook lane the primary listen route for your deployment, and do you expect the in-tree MCP `listen()` to run in the same deployment concurrently? It affects how we document the two listen paths.

**Change request (P1 — slot contention):** because this device-driven listen path is a deliberate, supported entry point — not a dead end — the recording-slot contention with MCP `listen()` is worth closing before merge. You already guard one direction: the `is_recording()` check makes a device `listen.start` defer to an active MCP `listen()` (nice). The reverse is still open — if a device-driven capture is active and an MCP `listen()` then acquires the slot, it takes over and the device-driven buffer is silently dropped (`esp32_client.py` ~567). Could you extend the same ownership guard to that direction — e.g. reserve the listen slot for the full device-driven start→stop window — so the two sources can't clobber each other? Since the whole point of the PR is a reliable external capture path, closing this makes it trustworthy when both sources can fire.

Device-safe either way, so no rush on timing — just flagging it as the one item to resolve before this merges.

kisaragi-mochi:
Separately, on the merge conflict here (this branch shows `CONFLICTING` on `CHANGELOG.md`, `esp32_client.py`, `gateway.py`): that's on our side, not yours. We landed the mDNS work (#229, #232) after you opened this PR, which moved the same gateway files. The rebase is ours to absorb — once your P1 fix above is in, we'll rebase this branch onto main on our end (`--force-with-lease`, your authorship intact). Nothing for you to do on the conflict itself.

maha0525:
Thanks for the careful read across #208 / #209 / #213–#215 / #216 — the integration framing matches the intent, that's exactly the audio I/O surface we're building toward.

## Positioning

External-hook is the primary listen route for our deployment. The MCP `listen()` tool is essentially unused day-to-day: a self-hosted gateway sits between stack-chan and our SAIVerse runtime, and SAIVerse owns capture orchestration (touch / wake / button → device-driven listen → hook → persona). We may keep `listen()` available for subagent / experimentation use, but it's the secondary path. Feel free to frame the docs that way — "device-driven external-hook is default, MCP `listen()` is the in-tree fallback."

## P1 (slot contention, reverse direction) — addressed in `227f74a`

Added the symmetric ownership guard inside `listen_and_transcribe` (`stt/orchestrator.py`), right after `session_id` retrieval and before `_begin_listen_motion` so we neither move the head nor send `listen.start` when we're going to decline. If `is_recording()` returns `True` when the MCP call enters the `lock_ctx`, it means a device-driven capture has the slot (since `lock_ctx` already serialises all MCP-side listen acquisitions and the device-driven path doesn't go through `lock_ctx`); the MCP call raises a `RuntimeError("...declined to avoid clobbering the active buffer")` rather than silently overwriting via `start_recording`.

- New test in `test_stt_orchestrator.py::test_pipeline_declines_when_device_driven_capture_active`: pre-acquires the slot with a device session id, calls `listen_and_transcribe`, asserts `RuntimeError` and that the pre-existing slot stays held (no clobber, no `listen.start` sent, no engine call).
- Full STT orchestrator suite stays green (27/27), `test_audio_stream.py` + `test_esp32_client.py` also pass (46/46).
- CHANGELOG entry for PR-F updated to describe both directions of the guard (and the `#TBD-F` placeholder is now `#209`).

## Codex pass on the rebased head

Re-checked the two Codex findings against the current head before deciding — both were already resolved by the 2026-05-21 follow-up commits in this branch, so flagging them here for the record rather than blocking on them:

- **P1 disconnect cleanup scope** (`esp32_client.py`): the `finally:` block at the bottom of `_handler` already guards with `self._device_driven_session_id == session_id` *and* `is_recording_session(session_id)` (the per-handler attribute is a session id string, not a bool, since `19a3...` "scope device-driven recording cleanup to the owning session").
- **P2 Ogg lacing** (`audio_input_hook.py`): `_packet_to_segments` (introduced in `e10c5b...`) handles both the >255 byte split and the trailing zero-length segment on exact-multiple-of-255 packets per RFC 3533 §6; `_build_ogg_page` validates the segment count.

So nothing left on the Codex side. Conflict resolution stays on your end as you mentioned — let me know if rebasing on top of this commit creates any friction.

kisaragi-mochi:
Rebased this branch onto current main and force-pushed (`--force-with-lease`), absorbing the CHANGELOG / `esp32_client.py` / `gateway.py` conflicts that came from the mDNS work (#229, #232) landing after this PR opened. That conflict was on our side, so the rebase is ours — your authorship is intact, all your commits are unchanged and just replayed onto main.

One maintainer fixup commit sits on top: the mDNS test fixture (`FakeEsp32` in `test_gateway.py`) didn't yet know about the `audio_hook_url=` / `audio_hook_token=` kwargs this branch adds to `ESP32Manager.start()`, so it raised `TypeError` after the rebase. The fixup just teaches the mock to accept and ignore those kwargs — no behavioural change, existing assertions untouched. Gateway test CI is green.

A local review on the rebased head also surfaced two minor (P2) items that are pre-existing PR logic, not rebase artifacts. Both are corner cases and device-safe, so they are not merge blockers — filed as follow-ups for whenever you want to pick them up:

- #237 — Ogg granule position on a mid-batch flush past the 255-segment page limit (only triggers on sustained very-high-bitrate VBR within a single ~3s page; normal 16 kHz mono STT audio never reaches it).
- #238 — recording-slot guard atomicity between the `is_recording` check and `start_recording` (only on a near-simultaneous device-driven capture and MCP `listen()`).

On the documentation framing question for the two listen paths — we'll sort that out separately so this PR stays focused on the feature itself; nothing needed from you there.

Thanks for the careful work here. We'll merge once CI is fully green.

PR #210, opened 2026-05-21
feat(avatar): dynamic AvatarSet transfer pipeline (HTTP fetch + AdoptOwnedBuffer)

maha0525:
## Summary

Adds a runtime-loadable avatar pipeline for stack-chan: a new `AvatarSet`
type that holds layered face / eyes / mouth frames in PSRAM, an HTTP
fetcher that streams a raw RGB565 payload from the gateway into PSRAM
without an intermediate copy, and a new `avatar_set_fetch` WebSocket
control protocol. Replaces the build-time `avatar_images.cc` table for
boards that opt into it. Matrix mode (90 frames, ~3.3 MB) is **not** in
this PR — that lives on a follow-up stacked PR.

## Why

The existing `avatar_images.cc` table is a baked-in code table — a new
avatar set requires reflashing. For stack-chan we want personas to swap
their face / eyes / mouth set at runtime, and we want to avoid bloating
the firmware partition with multiple sets.

This PR ships the transfer pipeline only. Default behavior is unchanged
until something on the firmware side calls into the new code path.

## How

- **Avatar payload format**: raw RGB565, packed in a fixed layout
  (face 6 frames × 320×240 × 2 bytes + eyes 3 + mouth 5 = 537,600 bytes
  for layered mode). No PNG decoder needed in firmware — the gateway
  delivers the pre-rendered bytes, sidestepping `CONFIG_LV_USE_PNG`
  partition pressure.
- **AvatarSetFetcher** (`firmware/main/boards/stackchan/avatar_set_fetcher.cc`):
  HTTP GET from the gateway's capture server, SHA256-validates against
  a checksum sent over WS, then hands the buffer to `AvatarSet` via
  `AdoptOwnedBuffer(uint8_t*, size_t)`. **No `memcpy`** — the staging
  buffer becomes the AvatarSet's storage, so the PSRAM peak is just
  `old_set + new_set` instead of `old_set + staging + new_set`. This
  matters for matrix mode (PR-E2), where the staging copy was
  overflowing 8 MB PSRAM and producing
  `Load: PSRAM allocation failed (size=3456000)`.
- **WS protocol**: gateway sends
  `{\"type\":\"avatar_set_fetch\",\"url\":...,\"checksum\":...,\"size\":...,\"mode\":...}`,
  device replies with `{\"type\":\"avatar_set_loaded\",\"ok\":...,\"checksum\":...,\"bytes_transferred\":...}`.
- **Expression-change deferral**: any `set_avatar_expression` arriving
  while a fetch is in progress is held until the new set is committed,
  so the display does not flicker between the old set and the partially
  loaded new set.
- **Gateway-side tooling**: new `load_avatar_set` MCP tool that takes
  a filesystem `archive_path` (not inline bytes — MCP JSON stays
  multi-MB-base64-free), stages it on the capture HTTP server with a
  short-lived Bearer-token-protected GET endpoint, and orchestrates
  the WS notify + reply wait.
- **Small `%zu` → `%u` fix** in two ESP_LOG sites. ESP-IDF's
  nano-printf cannot parse `%zu` (and crashes on it); using `%u` after
  casting to `unsigned int` is the safe pattern.
- **`Protocol::SendText` made public** so board-side code (the fetcher)
  can push the `avatar_set_loaded` reply over the active WS without
  re-implementing protocol plumbing. No behavior change for existing
  callers.

## Compatibility

- Default OFF. The new path is opt-in — boards that do not call into
  `AvatarSet` continue to use the existing `avatar_images.cc`
  approach. Stack-chan's existing layered face / eyes / mouth display
  works unchanged until something stages an AvatarSet over WS.
- The `avatar_set_fetch` message type is new; existing handlers ignore
  unknown types.
- New gateway endpoints are confined to a new HTTP route and a new MCP
  tool. Existing MCP tools (including `i2c_*`, `set_avatar*`,
  `set_leds`, motion tools) are unchanged.

## Test plan

- [x] Builds clean with ESP-IDF v5.5.2 (CI Docker image).
- [x] End-to-end verified on hardware (stack-chan board): a gateway-
      side `load_avatar_set` call with a layered-mode RGB565 file
      stages the bytes, the device fetches, validates, and commits the
      new set, and the LCD swaps to the new face / eyes / mouth on the
      next expression change.
- [x] Expression-change deferral verified — toggling
      `set_avatar_expression` while a fetch is in progress holds the
      change until completion, no partial-set flicker.
- [x] Repeated swaps work (no PSRAM leak across loads).

## Follow-up

PR-E2 (separate, will be opened off of this branch) adds **matrix
mode** (90 frames, ~3.3 MB) on top of this layered-mode foundation.
Kept separate so the matrix-mode rendering review can stay focused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -206,7 +206,6 @@ documented-only.
 
 ### Gateway
 
-<<<<<<< HEAD
 - Added: `send_pcm_audio(gateway, pcm, source_rate=...)` helper
   extracted from `synthesize_and_send`'s encode-and-push back-half.
   External producers — HTTP PCM bridges, sound-effect players,
@@ -218,8 +217,7 @@ documented-only.
   the existing `resample_pcm16_linear` helper, so producers at
   non-device rates work without manual resampling. Contributed via
   [PR #TBD-A1](https://github.com/kisaragi-mochi/stackchan-mcp/pull/TBD-A1).
-||||||| parent of bb37747 (docs(changelog): add Unreleased entries for PR-E1 (dynamic AvatarSet))
-=======
+
 - Added: `load_avatar_set` MCP tool + supporting HTTP staging /
   WebSocket fetch protocol for the firmware's dynamic AvatarSet
   pipeline (PR-E1 firmware side). The gateway stages a raw RGB565
@@ -229,7 +227,6 @@ documented-only.
   filesystem `archive_path` rather than inline bytes so the MCP JSON
   transport stays free of multi-MB base64 payloads. Contributed via
   [PR #210](https://github.com/kisaragi-mochi/stackchan-mcp/pull/210).
->>>>>>> bb37747 (docs(changelog): add Unreleased entries for PR-E1 (dynamic AvatarSet))
 
 - Fixed: `pip install stackchan-mcp[tts]` did not work out-of-the-box
   on Windows because `opuslib` calls `find_library("opus")`, which on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,33 @@ documented-only.
   and timing constants are unchanged. Contributed via
   [PR #206](https://github.com/kisaragi-mochi/stackchan-mcp/pull/206).
 
+- Added: dynamic avatar-set transfer pipeline. A new `AvatarSet`
+  scaffold (layered face / eyes / mouth, ~537 KB total in RGB565)
+  can be staged on the gateway and fetched into PSRAM at runtime via
+  a new `avatar_set_fetch` WebSocket protocol, replacing the old
+  build-time `avatar_images.cc` table. The fetch path uses a small
+  `AvatarSetFetcher` HTTP client that streams the raw payload
+  straight into a PSRAM staging buffer, validates SHA256, then hands
+  the buffer over to `AvatarSet::AdoptOwnedBuffer()` (an
+  ownership-transfer load, no memcpy). This caps the load-time
+  PSRAM peak at `old + new_size` instead of the pre-fix `old +
+  staging + new_size`, which was overshooting the 8 MB PSRAM budget
+  in matrix-mode (~3.3 MB) loads and producing `Load: PSRAM
+  allocation failed (size=3456000)`. Expression-change commands
+  arriving during a fetch are deferred until completion so the
+  display does not flicker between the old set and the partially
+  loaded new set. A small `%zu` → `%u` adjustment in two ESP_LOG
+  call-sites keeps nano-printf from crashing on those format
+  specifiers. Contributed via
+  [PR #210](https://github.com/kisaragi-mochi/stackchan-mcp/pull/210).
+
+- Fixed: `Protocol::SendText` was protected, so board-side code that
+  wants to push a board-initiated WS message (e.g. an `avatar_set_loaded`
+  reply from the AvatarSet fetcher) had no clean entry point.
+  Exposed as public; no behavioral change for existing callers.
+  Contributed via
+  [PR #210](https://github.com/kisaragi-mochi/stackchan-mcp/pull/210).
+
 - Fixed: a server-side close arriving between the WebSocket server
   hello and the main task resuming no longer cancels the reconnect
   timer that the per-socket `OnDisconnected` lambda just armed.
@@ -179,6 +206,7 @@ documented-only.
 
 ### Gateway
 
+<<<<<<< HEAD
 - Added: `send_pcm_audio(gateway, pcm, source_rate=...)` helper
   extracted from `synthesize_and_send`'s encode-and-push back-half.
   External producers — HTTP PCM bridges, sound-effect players,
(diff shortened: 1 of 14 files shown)


kisaragi-mochi:
Codex review against `main` after the rebase: 2 P2 findings on the introduced avatar pipeline. Both are functional / stability issues rather than release blockers, but each affects the pipeline's contract and is worth folding into the PR before merge.

### [P2] Matrix-mode loads accepted but never rendered

`firmware/main/boards/stackchan/stackchan.cc:3762` — when `load_avatar_set` is called with `mode="matrix"`, the firmware accepts the request and `AvatarSet::AdoptOwnedBuffer()` reports success, but the display lookup only consumes the dynamic set for `kLayered`. Matrix mode then falls through to the embedded static `avatar_images` frames, so the newly loaded 90-frame payload is never shown. From the gateway side the load looks successful while the device keeps rendering the old/static avatar.

The matrix-mode render plumbing is what PR #211 introduces, so one option is to land this PR as `kLayered`-only and let #211 handle the matrix render path. Another option is to either reject matrix loads with a clear error or wire the lookup here in this PR. Either way the current "accept then silently fall back" behaviour is the worst combination — it gives the gateway a false success signal.

### [P2] Avatar buffer swap races the LVGL display task

`firmware/main/boards/stackchan/avatar_set_fetcher.cc:118` — on a second or later dynamic avatar load while the avatar is visible, the worker task calls `AdoptOwnedBuffer()` without holding the LVGL/display lock. That method frees the previous PSRAM buffer and clears/repopulates descriptors that the existing `lv_image` may still be rendering. Stopping the avatar timers only prevents new `set_src` writes, not the LVGL display task from reading the current source, so repeated loads can race into a dangling/half-cleared descriptor and produce hard-fault or corrupted frames.

This one is best handled with a `DisplayLockGuard` around the `AdoptOwnedBuffer()` call (the same lock used by the rest of the firmware's LVGL boundary).

---

No urgency on either — first-load + single-load happy path is unaffected, so the PR isn't blocked from merge. Happy to wait for whichever direction you prefer (fix in this PR, carve-out into a follow-up, or land as-is with the matrix-mode behaviour documented as a known limitation pending #211).

maha0525:
Addressed both [P2] findings in this PR (fixed here rather than carved out).

**[P2] matrix loads accepted but never rendered** (`b2c2a6d`) — took the kLayered-only direction. `OnAvatarSetFetch` now rejects `mode="matrix"` with a `matrix_mode_unsupported` error instead of accepting it and silently falling back to the static tables. PR #211 lifts this guard when it wires the matrix lookup.

**On-device verified**: with this firmware flashed, sending a matrix avatar set now returns `{"ok": false, "error": "matrix_mode_unsupported"}` to the gateway — the same payload returned `{"ok": true}` (silent fall-back) before the fix. The caller now gets a deterministic failure instead of a false success.

**[P2] avatar buffer swap races the LVGL display task** (`ed06207`) — wrapped the `AdoptOwnedBuffer` swap in a `DisplayLockGuard` (null-display guarded for headless boards), following the existing LVGL-boundary locking pattern the rest of the board already uses. First-load / single-load happy path is unchanged. This one is **build-verified only**: the repeated-load race is inherently hard to prove absent on-device, so the fix rests on taking the same display lock the firmware's other LVGL writes already hold.

Both built clean against the stackchan target (ESP-IDF v5.5.4).

Note for the #211 cascade: since matrix is rejected here, #211 will drop the reject branch when it wires the matrix lookup. Happy to rebase #211 once this lands.

kisaragi-mochi:
Rebased the branch onto the latest main on our end (the conflict was just `mergeStateStatus: DIRTY` from main moving past, not actual content conflict — 13 commits replayed cleanly). Your `b2c2a6d` matrix-reject + `ed06207` DisplayLockGuard commits are preserved as `473e35b` / `ce5b60b`. Authorship intact, nothing for you to do.

Starting codex review on the rebased head now.

kisaragi-mochi:
@codex review

kisaragi-mochi:
Codex review on the rebased head found 3 items, but none block the happy path or risk the device. They're all polish-grade — a schema/firmware mismatch on matrix mode, a visual mismatch on dynamic mouth rendering, and a disconnect-time stall. Each is independently fixable.

We'll spin them out as follow-up issues on our end so this PR can land. If any of them happens to be in your way and you'd rather roll the fix into this PR (or a related branch in the stack), totally welcome — but it's not blocking on you, and we'll file the follow-ups regardless and link them back here once they're cut.

- **[P2] Gate `matrix` mode at the gateway until firmware accepts it** — `gateway/stackchan_mcp/stdio_server.py:739-744`. The MCP tool schema still advertises `mode: "matrix"` but firmware deterministically rejects it post-`b2c2a6d`. Visible to LLM callers as a "documented choice that always fails." Removing matrix from the gateway schema (or early-rejecting before staging the 3.4 MB payload) until PR #211 wires the matrix render restores the schema-reflects-what-works contract.
- **[P2] Route dynamic mouth frames through `AvatarSet`** — `firmware/main/boards/stackchan/stackchan.cc:4102-4103`. After `5ca84a8` restored the `MouthImageFor(shape)` static-table lookup, `SetMouthShape` resolves through the static table even after a layered set loads. The five new mouth frames in the transferred AvatarSet are never rendered, so post-load the device shows the new face + new eyes + static-table mouth.
- **[P3] Wake `_avatar_set_waiters` futures on disconnect** — `gateway/stackchan_mcp/esp32_client.py:70-73`. `disconnect()` already wakes the existing `_pending` futures, but the new avatar-set waiters aren't drained, so `load_avatar_set` stalls ~30 s on a known-gone connection instead of failing fast. A one-line addition mirroring the existing pattern.

The `b2c2a6d` matrix-reject + `ed06207` DisplayLockGuard fixes from the prior round look correct and addressed the earlier findings cleanly. Approving for merge once a final pass confirms CI is still green.

---

Separately — apologies for the chatter on this PR's timeline today. Our automated review workflow on this side fired a couple of extra comments (the rebase notice + a Codex bot re-trigger ping) that look noisier than the underlying change warrants. Nothing to act on, just flagging that the noise is on our end, not a sign of anything wrong with the PR.

kisaragi-mochi:
Follow-up issues filed for the three Codex review items, as promised:

- #226 — gate `matrix` mode at the gateway until firmware accepts it (P2)
- #227 — `SetMouthShape` bypasses `AvatarSet` routing → static mouth + new face/eyes mismatch (P2)
- #228 — `_avatar_set_waiters` not drained on disconnect → `load_avatar_set` stalls ~30s (P3)

All three are on our end now — feel free to roll any of them into this PR if it's convenient, otherwise they'll get picked up independently. PR #210 is good to merge from our side once CI confirms still green on the rebased head.

PR #212, opened 2026-05-21
refactor(tts): extract send_pcm_audio helper from synthesize_and_send

maha0525:
## Summary

Splits the encode-and-push back-half of `synthesize_and_send` into a
reusable `send_pcm_audio(gateway, pcm, source_rate=...)` function. The
existing `synthesize_and_send` now delegates to it after running the
engine, so all `say()` invocations keep working with no behaviour change.

## Why

External producers (HTTP PCM bridges, sound-effect players, alternative
voice stacks that already produce PCM) want to push pre-synthesised PCM
without registering a full `TTSEngine`. Today the only encode-and-push
path runs through `synthesize_and_send`, which is engine-coupled.

This PR is the first of a small stack:

- **#A1 (this PR)**: factor out `send_pcm_audio`.
- **#A2** (next): add `send_pcm_stream` for incremental push.
- **#A3** (after #A2): expose a `POST /pcm` endpoint that consumes the
  stream from an HTTP producer.
- **#A4** (after #A3): disable aiohttp's `client_max_size` cap on that
  endpoint for long audio.

Each later PR builds on the previous one. Splitting #A1 as the bottom of
the stack keeps the abstraction change reviewable in isolation.

## API

```python
async def send_pcm_audio(
    gateway: Gateway,
    pcm: bytes,
    *,
    source_rate: int = DEVICE_SAMPLE_RATE,  # 16_000
) -> None:
    \"\"\"Push raw mono PCM to the device, resampling if needed.\"\"\"
```

- `source_rate` lets producers at any rate hand over their PCM as-is;
  the function resamples to the device's 16 kHz mono using the existing
  `resample_pcm16_linear` helper.
- Validation, protocol gating, concurrency-lock, and disconnect-error
  translation match `synthesize_and_send`'s previous behaviour exactly.

## Test plan

- [x] All 14 pre-existing orchestrator tests still pass with no
      modification.
- [x] 14 new tests in `tests/test_send_pcm_audio.py` cover: happy path,
      resampling (8 kHz / 22.05 kHz / 44.1 kHz / 48 kHz inputs), input
      validation (empty, odd-length, wrong type), protocol gate
      (rejected when device not ready), concurrency lock (sequential
      enqueue), disconnect translation (mid-push WS close).

## Compatibility

Pure refactor at the call-site level — `synthesize_and_send`'s public
signature and behaviour are unchanged. The new helper is additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,18 @@ documented-only.
 
 ### Gateway
 
+- Added: `send_pcm_audio(gateway, pcm, source_rate=...)` helper
+  extracted from `synthesize_and_send`'s encode-and-push back-half.
+  External producers — HTTP PCM bridges, sound-effect players,
+  alternative voice stacks that already produce PCM — can now push
+  pre-synthesised audio to the device without registering a
+  `TTSEngine`. `synthesize_and_send` is unchanged from the caller's
+  perspective; it now delegates to `send_pcm_audio` after running
+  the engine. Resampling to the device's 16 kHz mono is done via
+  the existing `resample_pcm16_linear` helper, so producers at
+  non-device rates work without manual resampling. Contributed via
+  [PR #TBD-A1](https://github.com/kisaragi-mochi/stackchan-mcp/pull/TBD-A1).
+
 - Added: MCP tool surface for the firmware-side Grove Port A generic
   I2C bus introduced in
   [PR #196](https://github.com/kisaragi-mochi/stackchan-mcp/pull/196) —
--- a/gateway/stackchan_mcp/tts/__init__.py
+++ b/gateway/stackchan_mcp/tts/__init__.py
@@ -17,7 +17,7 @@ import logging
 from typing import Callable
 
 from .base import EngineRegistry, TTSEngine, get_registry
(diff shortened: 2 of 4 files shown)


kisaragi-mochi (approved):
Reviewed and LGTM. The encode-and-push extraction preserves `synthesize_and_send`'s observable return shape, and the new `send_pcm_audio` is well-defended (PCM / source_rate / gateway / device-connection / protocol-version validation, TTS-lock-protected start→frames→stop block, mid-stream `ConnectionError` translation). The 14 new tests cover the main regression surface (happy path, 4 resampling rates, input validation, protocol gate, concurrency, disconnect, pacing). CI green across all three checks.

A separate `codex review --base main` pass against the latest commit also produced no findings.

A small heads-up about the CHANGELOG: the Unreleased entry currently links to `PR #TBD-A1` — no need to push another commit on the branch for that, we'll fix it to `PR #212` on our side at merge time.

The branch shows as "behind main" because #223 just landed in the meantime — that's on our side, not yours. We'll click "Update branch" on this end before hitting merge, so nothing to do here.

Thanks for splitting the stack — the #A1 → #A4 progression keeps each step reviewable on its own, and the test-set on #A1 already covers the back-half regression surface for the later PRs.

PR #211, opened 2026-05-21
feat(firmware/avatar): AvatarSet matrix mode (full-frame rendering, ~3.3 MB PSRAM)

maha0525:
## Summary

Adds **matrix mode** to the AvatarSet pipeline introduced in PR #210
(layered face / eyes / mouth). Matrix mode pre-renders all
`face × eyes × mouth` combinations as full-screen RGB565 frames
(6 × 3 × 5 = 90 frames, ~3.3 MB) and indexes into the table at draw
time rather than compositing three regions per frame.

## Stacked on #210

This PR's `base` is `upstream/main`, but the commit history shows PR-E1
first (= #210) and then this one commit on top. Easiest review path: read
#210 first, then this PR (just `firmware/main/boards/stackchan/stackchan.cc`).

If #210 lands first, the diff here is a clean ~150-line patch.

## Why matrix mode

Layered mode (PR #210) composites face + eyes + mouth at draw time. For
artistic styles where the expression depends on cross-region interplay
(stylized characters with overlap, shadow, lighting per combination),
layered mode produces seams. Matrix mode lets the avatar artist deliver
a curated set of full-frame renders and skip composite logic entirely.

The trade-off is PSRAM: 3.3 MB vs 537 KB. For 8 MB PSRAM boards (like
stack-chan's CoreS3), this fits comfortably alongside the audio /
network / MCP allocations — PR #210's ownership-transfer fetch path
keeps the load-time peak at `old_set + new_set` rather than tripling.

## How

- `AvatarSet::mode_` carries which mode the loaded set is in
  (layered / matrix).
- `set_avatar_expression(name)` resolves to a `(face_idx, eyes_idx,
  mouth_idx)` triple; in matrix mode this indexes into the linear
  table at `face_idx × (eyes × mouth) + eyes_idx × mouth + mouth_idx`.
- Mode is fixed at `load_avatar_set` time. A persona swap that wants
  to switch modes goes through a full reload — there is no runtime
  mode toggle on a single loaded set.

## Compatibility

- Per-AvatarSet — boards that load a layered set get layered behavior,
  matrix set gets matrix. No firmware-wide flag.
- Default OFF in the sense that nothing uses matrix mode unless a
  persona explicitly stages a matrix set.

## Test plan

- [x] Builds clean (ESP-IDF v5.5.2, Docker CI image).
- [x] End-to-end on hardware: gateway `load_avatar_set` with a
      matrix-mode 3,456,000-byte payload completes successfully
[...]

PR #213, opened 2026-05-21
feat(tts): add send_pcm_stream for incremental PCM push

maha0525:
## Summary

Adds `send_pcm_stream(gateway, async_iter, source_rate=...)` as the
streaming counterpart to `send_pcm_audio` (PR #212). Consumes an async
iterable of PCM chunks, opus-encodes and pushes them frame-by-frame as
they arrive.

## Stacked on #212

This PR's commits sit on top of #212. The diff vs upstream/main therefore
includes #212's commit; the new code here is the second commit
(`send_pcm_stream` + tests).

## Why

`send_pcm_audio` (PR #212) buffers the entire PCM payload before pushing
the first opus frame. For long audio or genuinely streaming producers
(HTTP chunked uploads, real-time TTS engines that emit PCM as they
synthesise), that latency is wasted — playback could start as soon as
the first opus frame is encoded.

`send_pcm_stream` takes the same approach but pulls PCM chunks from an
`AsyncIterator[bytes]`. The protocol gate, concurrency lock, source-rate
resampling, and disconnect-error translation match `send_pcm_audio`.

## API

```python
async def send_pcm_stream(
    gateway: Gateway,
    pcm_stream: AsyncIterator[bytes],
    *,
    source_rate: int = DEVICE_SAMPLE_RATE,
) -> None:
```

Internally it accumulates partial PCM into 20 ms opus frames, encodes
and pushes complete frames immediately, and flushes the last partial
frame (zero-padded) when the iterator finishes.

## Test plan

- [x] 23 new tests in `tests/test_send_pcm_stream.py` cover: chunked
      happy path, source_rate resampling, frame-boundary partial chunk
      accumulation, EOS partial-frame padding, empty iterator,
      iterator-raises propagation, lock contention, protocol gating,
      and disconnect mid-stream.
- [x] `send_pcm_audio` (PR #212) and `synthesize_and_send` are
      unchanged and their tests continue to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,6 +258,15 @@ documented-only.
 
 - Added: the gateway now advertises `_stackchan-mcp._tcp.local.` over mDNS/DNS-SD by default so fresh firmware can discover the WebSocket endpoint on the local network. A new `--no-mdns` flag disables advertising.
 
+- Added: `send_pcm_stream(gateway, async_iter, source_rate=...)`
+  incremental variant of `send_pcm_audio`. Consumes an async
+  iterable of PCM chunks, opus-encodes and pushes them frame-by-frame
+  to the device as they arrive, with the same protocol gating /
+  concurrency lock as the buffered variant. Lets producers start
+  playback before the full audio is synthesised (HTTP streaming
+  upload, real-time TTS engines, etc.). Contributed via
+  [PR #213](https://github.com/kisaragi-mochi/stackchan-mcp/pull/213).
+
 - Added: `send_pcm_audio(gateway, pcm, source_rate=...)` helper
   extracted from `synthesize_and_send`'s encode-and-push back-half.
   External producers — HTTP PCM bridges, sound-effect players,
--- a/gateway/stackchan_mcp/tts/__init__.py
+++ b/gateway/stackchan_mcp/tts/__init__.py
@@ -17,7 +17,12 @@ import logging
 from typing import Callable
 
 from .base import EngineRegistry, TTSEngine, get_registry
-from .orchestrator import DEFAULT_VOICE, send_pcm_audio, synthesize_and_send
+from .orchestrator import (
+    DEFAULT_VOICE,
+    send_pcm_audio,
+    send_pcm_stream,
+    synthesize_and_send,
+)
 
 _logger = logging.getLogger(__name__)
 
@@ -52,5 +57,6 @@ __all__ = [
     "TTSEngine",
     "get_registry",
     "send_pcm_audio",
+    "send_pcm_stream",
     "synthesize_and_send",
 ]
--- a/gateway/stackchan_mcp/tts/orchestrator.py
+++ b/gateway/stackchan_mcp/tts/orchestrator.py
@@ -485,6 +485,21 @@ async def send_pcm_stream(
         DEVICE_SAMPLE_RATE * DEVICE_FRAME_DURATION_MS // 1000
     )
     bytes_per_frame = samples_per_frame * 2  # 16-bit
+    # Number of source-rate samples that produce exactly one device-rate
+    # opus frame after resampling. When the input is already at the device
+    # rate this equals ``samples_per_frame`` and the resample step is a
+    # no-op; otherwise we drain whole source-rate frames into the
+    # resampler, which avoids the rounding-error and odd-byte issues that
+    # per-chunk resampling has when transport chunk sizes are arbitrary.
+    src_samples_per_frame = (
+        source_rate * DEVICE_FRAME_DURATION_MS // 1000
+    )
+    if src_samples_per_frame <= 0:
+        raise RuntimeError(
+            f"source_rate {source_rate} is too low for "
+            f"{DEVICE_FRAME_DURATION_MS} ms frames"
+        )
+    bytes_per_src_frame = src_samples_per_frame * 2  # 16-bit
     encoder = opuslib.Encoder(
         DEVICE_SAMPLE_RATE, DEVICE_CHANNELS, opuslib.APPLICATION_VOIP
     )
@@ -494,6 +509,10 @@ async def send_pcm_stream(
 
     sent = 0
     push_error: ConnectionError | None = None
+    # ``buffer`` accumulates source-rate PCM bytes. Chunks may be odd-byte
+    # (HTTP chunked uploads can split a 16-bit sample across two transport
+    # chunks), so we accumulate raw bytes here and only resample / encode
+    # when the buffer holds at least one full source-rate frame's worth.
     buffer = bytearray()
 
     async def _push(opus_frame: bytes) -> bool:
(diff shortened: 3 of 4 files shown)


kisaragi-mochi:
Apologies — please disregard the rebase request above. The conflict situation on this stack is entirely on our side: we cut #212 first without coordinating with the rest of the stack, which broke the careful #A1 → #A4 sequence you laid out and forced the same three files (`CHANGELOG.md`, `gateway/stackchan_mcp/tts/__init__.py`, `gateway/stackchan_mcp/tts/orchestrator.py`) into conflict on #213, #214, and #215 at once. Asking you to clean up after our mishandling is not appropriate.

We'll handle the rebases on this end. Your branches will be pushed back with the #A1 commits dropped (since they're already squashed on main from #212), keeping the #A2 / #A3 / #A4 commits intact and preserving your authorship. You don't need to do anything — please leave the branches as they are.

We'll review the stack in order (#213 → #214 → #215) once the rebases land and CI is green on each. Two earlier Codex bot findings on this PR — P1 "Resample across stream, not per chunk" and P1 "Buffer sample alignment before per-chunk resampling" — show as outdated, so it looks like commit `0ceebea` ("buffer source-rate bytes before resampling") may have addressed them; we'll verify against the post-rebase head.

Thank you again for splitting this stack so cleanly. The cleanliness of the split is exactly what makes the rebase recoverable on our side; the only thing that needed coordinating was the merge order of #212 against the rest of the stack, and that's on us, not you. Sorry for the back-and-forth this caused — we'll be more careful with the merge order going forward.

kisaragi-mochi:
Rebased and force-pushed to all three branches (#213 / #214 / #215). The #A1 commits have been dropped (already on main from #212); the #A2 / #A3 / #A4 commits remain with your authorship intact. CI should kick off automatically on each branch.

Once CI is green on each, we'll review the stack in order (#213 → #214 → #215). You don't need to do anything in the meantime — apologies again for the back-and-forth and thank you for your patience.

kisaragi-mochi (commented):
Codex review on the rebase head flagged one issue worth addressing before merge:

**[P1] Pacing relative to actual send time** (`gateway/stackchan_mcp/tts/orchestrator.py:521-530`)

When the async PCM producer pauses before yielding (= exactly the scenario this helper supports — HTTP chunked uploads, streaming TTS with synthesis jitter), `next_send_time` stays anchored to just after `tts.start`. A later chunk with many frames bursts them back-to-back until the schedule catches up, which can exceed the firmware decode queue (~40 packets) and silently drop audio.

Since this hits the PR's own intended use cases, it warrants a fix rather than carve-out. Suggested approach is up to you — anchoring `next_send_time` to the actual last-send time, or per-frame `monotonic()` re-check, both should work. Happy to discuss trade-offs.

Once a fix lands and CI is green, we'll do a final review pass.

kisaragi-mochi:
Stack-wide status update (#213 → #214 → #215):

- **#213 (this PR)**: one P1 pacing finding to address before merge (see codex review comment above).
- **#214**: one description ↔ impl mismatch on the `/pcm?rate=` query-string fallback, small either-way fix (see PR #214 codex review).
- **#215**: one scope item carved out as a follow-up issue, accepting as-is (see PR #215 codex review).

Once #213's pacing fix lands, the stack will need a cascade rebase on this end (since #214 / #215 share the rebased #A2 base), and then we'll do a final review pass and approve in order #213 → #214 → #215.

Take your time — no rush on the timing.

maha0525:
Thanks for handling the stack rebase end-to-end — landing the rebased #A1 cleanly and keeping authorship intact on the rest was the right call given the merge-order mishap, and the apology was more than needed. No hard feelings.

## P1 pacing — addressed in `2259036`

Anchored the schedule advance to the later of the previous target (preserves the 20 ms cadence when keeping up) and the actual current time after the send (re-anchors to real time when the upstream producer paused). With this, a post-pause multi-frame chunk paces at `frame_period_s` per send instead of bursting back-to-back.

```python
# Before:
sent += 1
next_send_time += frame_period_s

# After:
sent += 1
next_send_time = max(next_send_time, loop.time()) + frame_period_s
```

`send_pcm_audio` (PR #212) is not affected — its producer is the synchronous encoder loop with no upstream pause between consecutive `_push` calls.

Tests:

- New `test_stream_re_paces_after_producer_pause` yields one frame, awaits `asyncio.sleep(5 * frame_period_s)`, then yields a 5-frame chunk, and asserts each consecutive post-pause send is at least one frame period apart (with a 5 ms scheduling-jitter tolerance). Verified the test **fails on the pre-fix code** (good regression coverage) and passes after the fix.
- Full `test_send_pcm_stream.py` suite: 18/18 (17 pre-existing + 1 new). Wider sweep (`test_send_pcm_audio` + `test_audio_stream` + `test_orchestrator`): 53/53.

## Codex pass on the rebased head

Re-checked the two older Codex P1s against the current head before deciding — both already resolved by `28beb7f` ("buffer source-rate bytes before resampling"), as you anticipated:

- **P1 "Resample across stream, not per chunk"**: the drain loop now accumulates source-rate bytes in `buffer` and resamples per whole-source-frame, so each whole source frame produces exactly `samples_per_frame` device samples and the per-chunk rounding error is gone.
- **P1 "Buffer sample alignment before per-chunk resampling"**: `buffer.extend(chunk)` accepts odd-byte chunks before any resample call, so trailing partial samples carry across chunk boundaries.

The post-rebase Codex pass should clear both.

## Small piggy-back changes in the same commit

Surfacing these explicitly since they widen the diff a bit beyond the pacing line:

- `test_stream_translates_encoder_error` switched from raw `fake_opuslib.Encoder.encode = boom_encode` to `monkeypatch.setattr(fake_opuslib.Encoder, "encode", boom_encode)`. `_FakeOpusEncoder` is a module-level class so the raw assignment leaked into the new pacing test under default collection order; this also prevents future tests from tripping over the same hidden coupling.
- CHANGELOG `#TBD-A2` placeholder → `#213`.

Happy to split either out if you prefer separate commits — they were small enough that I bundled them, but the pacing fix is the substantive part.

## Stack cascade

Acknowledging that this triggers the cascade rebase on #214 / #215 as you noted. Take whatever pace works on your end — no rush.

kisaragi-mochi:
Rebased onto the latest `main` and force-pushed to `pr-a2-send-pcm-stream`. The only conflict was in `CHANGELOG.md`: #211 (avatar matrix mode) landed first and added Gateway entries in the same `[Unreleased]` section. That's on our side, not yours, so we resolved it on this end — all three entries are kept, and your `send_pcm_stream` entry now references #213 instead of the `TBD-A2` placeholder. Your commits are unchanged and your authorship is intact, so there's nothing to do here.

While reviewing the rebased branch locally, one minor robustness note came up. Rather than expanding this PR's scope, we filed it as a separate low-priority tracking issue (#236):

`send_pcm_stream` doesn't type-check `source_rate` the way `send_pcm_audio` does. In practice this has no impact on any supported path — `POST /pcm` already coerces the `X-Sample-Rate` header via `int()` and rejects malformed values before dispatching here, and the in-process TTS path uses the buffered helper. It would only surface if a future integration called `send_pcm_stream` directly with an untyped `source_rate`. The recommended convention (external/real-time producers route through `/pcm`) keeps that covered, so it's just a note for later — no action needed on this PR.

All CI checks pass — gateway test, CHANGELOG check, firmware build, and windows-smoke are green.

PR #214, opened 2026-05-21
feat(capture_server): add POST /pcm endpoint for external PCM input

maha0525:
## Summary

Adds a `POST /pcm` HTTP endpoint to the capture server that lets external
producers push PCM directly to the device. Internally streams chunks
through `send_pcm_stream` (PR #213) so playback starts as soon as the
first opus frame is encoded.

## Stacked on #213 (and #212)

Diff vs upstream/main includes the prior PRs in the stack. The new code
here is the third commit (HTTP endpoint + tests).

## Why

The TTS-engine path is engine-coupled. The `send_pcm_audio` / `send_pcm_stream`
helpers (PR #212 / #213) let in-process producers push PCM, but external
producers — sound-effect players, browser bridges, alternative voice
stacks running in their own processes — need a network-reachable entry
point.

This PR adds the smallest plausible such entry: an HTTP POST that takes
PCM and pipes it through the existing helpers.

## API

```
POST /pcm
Authorization: Bearer <STACKCHAN_PCM_TOKEN>
Content-Type: application/octet-stream
Transfer-Encoding: chunked
X-Sample-Rate: <int>            (source rate, resampled to 16 kHz on device)
X-Channels: 1                   (only mono supported)
X-Message-Id: <optional>        (opaque correlation id for logging)

<raw signed-16-bit LE mono PCM bytes>
```

- Bearer-token auth (matches the other capture endpoints). The
  `STACKCHAN_PCM_TOKEN` env var sets the expected token; falls back to
  `STACKCHAN_TOKEN` / `BEARER_TOKEN` for single-token local
  development.
- The source sample rate is carried in the `X-Sample-Rate` header. The
  handler resamples to the device's 16 kHz before Opus encoding.
- Body is streamed — chunks flow through `send_pcm_stream` as they
  arrive, so first-opus-frame latency is bounded by network RTT + one
  frame's worth of PCM (~20 ms), not the full payload size.

## Compatibility

- New endpoint, additive. Existing capture endpoints unchanged.
- Default OFF in the sense that nothing pushes to `/pcm` unless an
  external producer is configured to do so.

## Test plan

- [x] New endpoint tests cover: happy path, auth-rejection, missing
      rate, malformed Content-Type, stream truncation, and concurrent-
      push lock contention.
- [x] PRs #212 / #213 tests still pass.
- [x] End-to-end verified: a Python script that POSTs a WAV's PCM body
      to `/pcm` plays cleanly on the device.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -382,6 +382,16 @@ documented-only.
 
 - Added: the gateway now advertises `_stackchan-mcp._tcp.local.` over mDNS/DNS-SD by default so fresh firmware can discover the WebSocket endpoint on the local network. A new `--no-mdns` flag disables advertising.
 
+- Added: `POST /pcm` HTTP endpoint on the capture server that consumes
+  external PCM uploads and pipes them through `send_pcm_stream` to the
+  device. Lets non-MCP producers (sound-effect players, alternative TTS
+  stacks, browser bridges) hand audio to stack-chan over plain HTTP
+  without registering a `TTSEngine`. Accepts raw PCM with the source
+  sample rate carried in an `X-Sample-Rate` header, requires
+  Bearer-token auth, streams chunks to the device as they arrive so
+  latency stays low for long uploads. Contributed via
+  [PR #214](https://github.com/kisaragi-mochi/stackchan-mcp/pull/214).
+
 - Added: `send_pcm_stream(gateway, async_iter, source_rate=...)`
   incremental variant of `send_pcm_audio`. Consumes an async
   iterable of PCM chunks, opus-encodes and pushes them frame-by-frame
--- a/gateway/stackchan_mcp/capture_server.py
+++ b/gateway/stackchan_mcp/capture_server.py
@@ -1,11 +1,46 @@
-"""HTTP capture server for receiving photos from ESP32.
+"""HTTP capture server for receiving photos from ESP32 and PCM from external producers.
 
-ESP32's camera.Explain() POSTs multipart/form-data with:
-- field 'question' (text)
-- field 'file' (camera.jpg, JPEG image)
+Two POST endpoints share this server:
 
-This server saves the JPEG and returns the file path so MCP client
-can view the image via the Read tool.
+- ``POST /capture``: ESP32's camera.Explain() uploads JPEG photos as
+  multipart/form-data (fields: ``question`` text + ``file`` JPEG).
+  Authenticated via ``CAPTURE_TOKEN_KEY`` (the gateway's ``vision_token``).
+  The server saves the JPEG to ``~/.stackchan/captures/`` and returns the
+  file path so the MCP client can view the image via the Read tool.
+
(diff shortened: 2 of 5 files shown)


kisaragi-mochi (commented):
Codex review on the incremental diff (base=pr-a2-send-pcm-stream) flagged one mismatch:

**[docs] Documented `/pcm?rate=...` query-string vs `X-Sample-Rate` header-only impl** (`gateway/stackchan_mcp/capture_server.py:170-172`)

The PR description and CHANGELOG advertise `POST /pcm?rate=32000` as a fallback for clients that can't set Content-Type parameters, but the handler only reads `X-Sample-Rate`. A client following the description gets a 400 before audio streaming starts.

Two options, either works:
- Read the `rate` query parameter as a fallback alongside `X-Sample-Rate`.
- Update the description to header-only (drop the query-string mention).

Either way is a small change. Up to you which direction fits the design.

Other than that, the endpoint is well-structured and tests are thorough. Looks good once this is reconciled.

kisaragi-mochi:
Thanks for the stack — and apologies again for the churn on our side (we landed #210 / #211 / #213 and the mDNS work ahead of this, which is what put this branch behind).

We'll absorb the rebase on this end: the conflicts are all ours (the mDNS `start()` signature from #229, and #210's avatar handlers landing in `capture_server.py` next to your `/pcm` handler — all additive). We'll re-base your `/pcm` work onto the current main, which now includes #213 with its post-merge pacing fix. Nothing to do on your side there.

One small thing while we're in there: the handler reads the sample rate from the `X-Sample-Rate` header (and the module docstring documents it that way), but the CHANGELOG entry describes it as a query-string parameter — looks like the changelog wording just drifted from the implementation. The header is clearly the intended design, so we'll align the CHANGELOG wording to the header as part of the rebase commit. Flagging it here so you're aware — let us know if you actually meant query-string, otherwise we'll keep the header form.

maha0525:
Thanks — and no worries at all about the churn, that's the cost of a stack sitting open while main moves. Much appreciated that you're absorbing the rebase on your end.

On the sample rate: yes, the header (`X-Sample-Rate`) is the intended design. The handler, the module docstring, and the test suite all read/assert the header form consistently — the query-string wording in the CHANGELOG is just drift, not a deliberate alternative. (Same drift snuck into this PR's own description, which mentions an `audio/L16; rate=` / `?rate=` form that the code never actually implements — please disregard that too.)

So please keep the header form and align the CHANGELOG to it. Nothing query-string-related was intended. Thanks again!

kisaragi-mochi:
Rebased onto current main (7e9576f) and force-pushed to `pr-a3-pcm-endpoint`. The pr-a2 commits dropped cleanly (already on main via #213), so only the three pr-a3 commits land on top of main.

The two scope items you flagged in your earlier comment are addressed:

- CHANGELOG entry now uses the `X-Sample-Rate` header form and references PR #214 (was placeholder `TBD-A3`).
- PR description: removed the `audio/L16; rate=` / `?rate=` mentions per your note.

CI is running now; once green, this should be ready for the next review pass.

Issue #178, opened 2026-05-18
feat(gateway): MCP-over-HTTP transport + ESP32 command queue for transparent multi-client sharing


kisaragi-mochi:
## Problem

The current gateway is a stdio MCP server: each MCP client that registers `stackchan-mcp` spawns its own gateway process. Because the ESP32 firmware (xiaozhi-esp32 base) accepts only one WebSocket, concurrent gateway processes contend for the single device WS, and at most one client can actually control the device at a time.

Even with the tactical ownership-lock work in #177 (which arbitrates *who* gets the WS), the underlying single-client serialization is fundamentally limiting for realistic workloads:

- Multi-agent setups (one orchestrator, several worker sessions) cannot share the device.
- Long-running automations (cron, daemons, observability dashboards) cannot coexist with interactive use.
- The session that "owns" the device blocks every other consumer until it explicitly releases.

## Real-device evidence (2026-05-19) — priority elevation rationale

Updated to `priority/medium` based on the following real-device verification.

### Independent controlled experiment

Two dummy MCP server variants (`bind` / `nobind`) were exercised under port 8765 contention. With port 8765 already occupied by an existing gateway:

| Variant | Result |
|---|---|
| `bind` (binds 8765 unconditionally) | `[Errno 48] Address already in use` on startup, no MCP `initialize` response — client sees a dead stdio gateway |
| `nobind` (skips the bind) | Successful `initialize` + `tools/list` exchange — the failure mode is isolated to the bind step |

This confirms that the **single failure point is port 8765 contention**, and that the MCP stdio transport itself is healthy. The structural fix scoped by this Issue (HTTP transport + single shared gateway) eliminates the contention by design — no future "second gateway tries to bind 8765" path exists.

### Frequency revised

The contention was originally categorized as a "rare event triggered only during alternate-agent handoff" scenario (hence `priority/low`). Real-device observation on 2026-05-19 shows the contention also occurs during **routine multi-session operation** — e.g. opening a second MCP client session that registers `stackchan-mcp` while the first session's gateway is still alive.

One concrete incident on 2026-05-19:

- 17:36 — second session launched, port 8765 already held by the first session's gateway → second session's gateway spawn failed → expected MCP tools (23) absent from the second session's tool list → roughly 4 hours of downstream investigation lost on the second session's primary task before the cause was isolated.

The handoff-only mental model understates the real impact. Routine multi-session operation is the more common trigger.

### Decision

[...]

Issue #177, opened 2026-05-18
feat(gateway): WebSocket ownership lock — single-client coordination for shared ESP32 device


kisaragi-mochi:
## Problem

The ESP32 firmware (xiaozhi-esp32 base) accepts **only one WebSocket connection** at a time. The stackchan-mcp gateway runs as a stdio MCP server, so a separate gateway process gets spawned for each MCP client that registers `stackchan-mcp` (e.g. one per Claude session, one per Codex session, one per external agent).

When two such gateways start up simultaneously, both try to claim the ESP32 WebSocket. The ESP32 drops the previous connection in favor of the new one, leaving the first client with a "ghost" gateway: the stdio MCP layer is still alive, but every device-facing tool call hits a disconnected WS and either fails silently or hangs.

Concrete failure pattern observed in practice:

- Two MCP clients on the same host each register `stackchan-mcp` and spawn their own gateway processes.
- One operator manually kills one of the gateway processes (to "clean up").
- Both sessions lose stackchan-mcp tools simultaneously, because the kill cascades through the shared ESP32 WS.

There is currently no signal at all to a second gateway that the device is already owned.

## Real-device evidence (2026-05-19) — tactical urgency

Independent controlled experiment (`bind` / `nobind` dummy MCP server pair) confirmed port 8765 contention as the single failure mode of the current stdio gateway design. Full experiment results and methodology are documented in #178's "Real-device evidence" section.

Routine multi-session operation triggers the contention with measurable downstream impact. One concrete incident on 2026-05-19:

- 17:36 — second MCP client session launched, port 8765 already held by the first session's gateway → second session's gateway spawn failed silently → roughly 4 hours of downstream investigation lost on the second session's primary task before the cause was isolated.

#178 (HTTP transport + single shared gateway) is the structural fix but requires #169 to land first and is a larger refactor. This Issue's tactical ownership-lock fix is the **bridge** during the #178 design and implementation window: it converts the silent-failure mode into a deterministic, diagnosable refuse / queue / preempt outcome.

`priority/medium` is preserved with the understanding that this Issue stays the recommended near-term landing target even after #178 design work begins, and is gracefully superseded once #178 ships.

## Goal

Add gateway-side **ownership coordination** so that:

1. The gateway tracks which client currently owns the ESP32 WebSocket connection.
2. A second gateway instance starting up can detect an existing owner and choose a defined behavior (refuse / queue / preempt) instead of silently breaking the first one.
[...]

PR #215, opened 2026-05-21
fix(gateway): disable aiohttp client_max_size cap on capture app

maha0525:
## Summary

aiohttp's `web.Application` defaults `client_max_size` to 1 MiB, applied
to the cumulative body of every incoming request. `POST /pcm` (PR #214)
is intentionally a streaming receiver — external producers chunk
arbitrarily long PCM utterances through the same long-lived HTTP request
— and the byte total of a long utterance routinely exceeds 1 MiB.

When the cap is hit, aiohttp aborts the request body read mid-stream
without surfacing a clean error to the handler, so the device hears a
truncated utterance and the producer gets no indication that anything
went wrong.

This PR sets `client_max_size=0` on the capture `web.Application`, which
disables the cap entirely. The PCM endpoint's own handler still bounds
memory by streaming chunks straight into `send_pcm_stream`'s opus
encoder, so memory pressure is per-frame regardless of body size.

## Stacked on #214

Diff vs upstream/main shows the full A1→A4 stack. The change in this PR
is a single-line config flag flip on the capture web.Application.

## Why disable, not just raise?

A bounded cap would require knowing the maximum reasonable utterance
size, and there is no principled answer — \"live mix from a browser\"
and \"morning briefing read aloud\" both want unbounded length. The
streaming endpoint architecture already enforces memory boundedness
per-frame, so a body-size cap adds no safety, only a footgun.

## Compatibility

- Other endpoints on the same app (image upload, vision capture) impose
  their own size checks in their handlers — they don't rely on
  `client_max_size` for memory safety.
- No new public surface. Existing in-process callers see no behaviour
  change.

## Test plan

- [x] Verified locally: a multi-minute TTS audio (~3 MiB body)
      streams through `/pcm` without truncation. Before this PR the
      same audio cut off at ~1 MiB.
- [x] PRs #212 / #213 / #214 tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -392,6 +392,22 @@ documented-only.
   latency stays low for long uploads. Contributed via
   [PR #214](https://github.com/kisaragi-mochi/stackchan-mcp/pull/214).
 
+- Fixed: aiohttp's default 1 MiB `client_max_size` cap on the capture
+  `web.Application` was aborting `POST /pcm` requests mid-stream for
+  long PCM uploads (multi-minute TTS, live mixes, persistent audio
+  feeds). The cap is meaningful only as a body-size limit for buffered
+  request bodies, and `/pcm` is intentionally a streaming endpoint —
+  the byte total of an utterance is bounded by the producer, not by
+  what fits in a single buffer. Setting `client_max_size=0` on the
+  capture app disables the cap. A per-route 8 MiB cap
+  (`CAPTURE_MAX_BYTES`) is added to `/capture` to retain a body-size
+  guard there, since that endpoint streams JPEG uploads to disk;
+  oversized uploads are rejected with `413 Payload Too Large` (with
+  partial files cleaned up). The PCM route remains uncapped because
+  external producers intentionally stream arbitrarily long audio
+  there. Contributed via
+  [PR #215](https://github.com/kisaragi-mochi/stackchan-mcp/pull/215).
+
 - Added: `send_pcm_stream(gateway, async_iter, source_rate=...)`
   incremental variant of `send_pcm_audio`. Consumes an async
   iterable of PCM chunks, opus-encodes and pushes them frame-by-frame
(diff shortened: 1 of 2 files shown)


kisaragi-mochi (commented):
Codex review on the incremental diff (base=pr-a3-pcm-endpoint) flagged one scope item that's better carved out than blocking this PR:

**[scope] Chunked-multipart non-`file` parts on `/capture` aren't bounded** (`gateway/stackchan_mcp/capture_server.py:342`)

The `33660d5` per-route cap handles `Content-Length` and the `file` part, which is the common path. Chunked multipart with no `Content-Length` can still pass a `question` or other non-`file` part through `await part.read()` unbounded. In the current single-user / single-WS deployment this is a narrow exposure; broader DoS hardening lines up better with the multi-client roadmap (#177 ownership lock / #178 MCP-over-HTTP queue) than with this PR's scope.

Proposing to accept this PR as-is and carve out a follow-up issue for "bounded multipart parsing on /capture" that gets folded into the multi-client work. Happy to file that issue on our end after merge.

The rest of the change (disabling the global cap for `/pcm` streaming + per-route cap on `/capture`) reads correctly and tests pass.

kisaragi-mochi:
Followed up by rebasing `pr-a4-client-max-size` onto current main as well (stack-rebase absorb continuation). The 6 pr-a2 / pr-a3 commits dropped cleanly (already on main via #214) and the 3 pr-a4 commits land on top.

CHANGELOG entry was reformatted: the `TBD-A4` PR placeholder is now `#215`, and the entry now covers both the `client_max_size` disable and the new per-route 8 MiB cap on `/capture` (the cap was missing from the original CHANGELOG commit, so I folded it into the same Unreleased item).

Author / `Co-Authored-By` trailers preserved as-is. CI is running now.

kisaragi-mochi:
Filing the carve-out issue from the review now.

PR #216, opened 2026-05-21
fix(firmware): skip xiaozhi OTA check to preserve user-configured WebSocket gateway URL

maha0525:
## Summary

`Application::ActivationTask()` calls `Ota::CheckVersion()` on every
boot, which POSTs the device descriptor to the upstream xiaozhi OTA
endpoint (`api.tenclass.net` by default). The response handler in
`ota.cc::CheckVersion()` then **writes every `websocket` and `mqtt`
key from the response back into the corresponding NVS namespaces** —
**overwriting** the gateway URL the user just saved through the WiFi
config UI's \"Advanced > WebSocket Gateway URL\" field.

The result is that a freshly configured device pointed at the user's
self-hosted gateway falls back to the xiaozhi-cloud URL after the next
reboot.

## Why this is safe to skip in this fork

stackchan-mcp is explicitly designed to speak to a self-hosted gateway,
not to xiaozhi-cloud:

- `InitializeProtocol()` already forces `WebsocketProtocol` — there is
  no MQTT path.
- `websocket_protocol.cc` has its own NVS-backed URL field with
  primary+fallback support, set through the WiFi config UI.
- The OTA-config payload that `CheckVersion()` consumes was a feature
  of upstream xiaozhi for managing fleets of cloud-attached devices.
  In this fork, the only side effect of `CheckVersion()` is the
  unintended overwrite.

This PR makes `CheckVersion()` early-return with a single log line.
Activation flow continues without it.

## Compatibility concerns

I want to flag this for the maintainer because it intentionally severs
contact with the xiaozhi-cloud OTA flow:

- If there's a future intent to support xiaozhi-cloud device management
  alongside the self-hosted gateway, this PR would conflict with that
  direction. I would understand a request to put this behind a Kconfig
  flag (`CONFIG_STACKCHAN_SKIP_XIAOZHI_OTA` default y) rather than
  hard-disabling it.
- Firmware version reporting via the OTA endpoint is also lost, but
  the firmware release flow in this repo doesn't rely on that endpoint.

I'm happy to gate this behind Kconfig if you'd prefer — the simplest
form (unconditional skip) is what I've been running on our hardware
since the issue started biting users who reconfigured their gateway URL
and then rebooted.

## Test plan

- [x] Built with ESP-IDF v5.5.2 (CI Docker image).
- [x] Flashed to hardware. User-configured WebSocket Gateway URL
      survives reboot (verified by checking
      `websocket://...://nvs.read?ns=websocket&key=url` after a
      reboot — previously was reset to `api.tenclass.net`).
- [x] Device still completes Activation and connects to the configured
      gateway normally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,22 @@ documented-only.
 
 ### Firmware
 
+- Fixed: every boot, `Application::ActivationTask()` called
+  `Ota::CheckVersion()`, which posts the device descriptor to the
+  upstream xiaozhi OTA endpoint (`api.tenclass.net` by default). The
+  response handler in `ota.cc::CheckVersion()` then writes every
+  `websocket` and `mqtt` key from the response back into NVS,
+  overwriting the gateway URL the user just saved through the WiFi
+  config UI's "Advanced > WebSocket Gateway URL" field. stackchan-mcp
+  is designed to speak to a self-hosted gateway —
+  `InitializeProtocol()` already forces `WebsocketProtocol`, and
+  `websocket_protocol.cc` has its own NVS-backed URL field with
+  primary+fallback support — so the xiaozhi OTA-config payload was
+  never meant to be consumed in this fork. `CheckVersion()` is now
+  skipped (early-returns with a single log line) so the user-
+  configured gateway URL survives reboot. Contributed via
+  [PR #TBD-B](https://github.com/kisaragi-mochi/stackchan-mcp/pull/TBD-B).
+
 - Fixed: WebSocket candidate fallback is now fail-fast when a server
   hello is malformed (missing/non-string `transport`, missing/empty
   `session_id`, or unsupported `transport`). A new
--- a/firmware/main/application.cc
+++ b/firmware/main/application.cc
@@ -324,11 +324,22 @@ void Application::ActivationTask() {
     // Create OTA object for activation process
     ota_ = std::make_unique<Ota>();
 
-    // Check for new assets version
+    // Check for new assets version (uses NVS only, does not contact any
+    // external server)
     CheckAssetsVersion();
 
-    // Check for new firmware version
-    CheckNewVersion();
+    // NOTE: The legacy xiaozhi-cloud OTA check is intentionally skipped.
+    // ota.cc::CheckVersion() has a side effect of overwriting the NVS
+    // "websocket" and "mqtt" namespaces with whatever the OTA response
+    // carries, which clobbers the gateway URL the user sets via the WiFi
+    // config UI (Advanced > WebSocket Gateway URL). The stackchan-mcp
+    // firmware always speaks to a custom gateway directly (see
+    // InitializeProtocol() below and websocket_protocol.cc), so the
+    // OTA-config payload is never consumed; skipping the call is the
+    // simplest way to keep the user-configured URL intact across reboots.
+    // current_version_ is populated in the Ota constructor from the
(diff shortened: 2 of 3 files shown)


kisaragi-mochi (commented):
Codex review (base=main) flagged one item that lines up with the contingency you already called out in the PR description:

**[P2] Preserve self-hosted custom OTA path** (`firmware/main/application.cc:331-351`)

Replacing `CheckNewVersion()` with only `MarkCurrentVersionValid()` also disables the legitimate self-hosted firmware-update flow for users who configure `wifi/ota_url` through the WiFi config UI — not just the xiaozhi-cloud clobbering you targeted.

You already mentioned a Kconfig flag (`CONFIG_STACKCHAN_SKIP_XIAOZHI_OTA` default y) as an alternative in the PR description — we'd like to take you up on that. Keeping the OTA path behind a Kconfig with sensible defaults preserves self-hosted OTA users while shipping the URL-preservation fix everyone benefits from.

The rest of the change reads well: the `current_version_` populate in the `Ota` ctor is a nice tightening, the `MarkCurrentVersionValid()` preservation is correct, and the surrounding comment is thorough.

When you push the Kconfig fix, a rebase onto the latest main while you're at it would be helpful — the CHANGELOG.md `[Unreleased]` section has moved since this branch was opened (PR #212 landed in the meantime), so the fix can ride on top of clean main. If folding the rebase in is awkward, we can resolve the CHANGELOG conflict on our end at merge time instead.

kisaragi-mochi:
Update on the previous comment — apologies for the slow correction.

Looking again at the existing fix from PR #119 (merged 2026-05-14), the user-facing problem this PR targets (= self-hosted WebSocket URL being overwritten by OTA response) is already resolved upstream:

- PR #119 added `CONFIG_DISABLE_OTA_WEBSOCKET_CONFIG` (default y), which gates the `websocket`/`mqtt` NVS-writeback inside `Ota::CheckVersion()`. The URL the user saves through the WiFi config UI now survives reboot regardless of what the OTA endpoint returns.

If your motivation here is the same URL-preservation issue, this PR may actually be redundant with #119 — could you confirm whether that's the case? If so we can close this with thanks.

If your motivation is broader (e.g. skipping the OTA round-trip itself for bandwidth / latency / privacy reasons even when the NVS-writeback is already gated), that's a separate axis from #119 and worth a fresh conversation. In that case the codex finding ([P2] Preserve self-hosted custom OTA path) would still apply, and naming alignment with the existing `CONFIG_DISABLE_OTA_WEBSOCKET_CONFIG` would help keep the Kconfig surface tidy.

Sorry for not surfacing #119 in the first version of the review above — that was on us. Either direction works; we're flexible.

maha0525:
Confirmed — this PR is redundant with #119, so closing it.

You were right: the motivation here was exactly the URL-preservation issue (keeping the self-hosted `ws://<lan-ip>` from being clobbered by the OTA response). #119's `CONFIG_DISABLE_OTA_WEBSOCKET_CONFIG` (default `y`) already gates the `websocket`-section NVS write-back, which fully covers that.

The only delta in this PR was skipping the OTA round-trip itself rather than just gating the write-back, but that was an implementation choice in service of URL preservation, not a separate goal — and as the codex [P2] note flagged, it has the downside of disabling the legitimate self-hosted OTA path. With the user-facing problem already solved upstream, there's no reason to take on that tradeoff.

Thanks for surfacing #119, and no worries about the timing of the correction. Closing with thanks.

PR #217, opened 2026-05-21
fix(gateway): bundle libopus.dll for Windows + PATH setup

maha0525:
## Summary

Make `pip install stackchan-mcp[tts]` work out-of-the-box on Windows
x64 by publishing a platform-specific `*-win_amd64.whl` that bundles
`opus.dll` built from upstream Opus source by the publish workflow.
Non-Windows wheels and the sdist stay clean (no bundled binary, they
rely on system `libopus`); non-x64 Windows sdist installs are guarded
by an architecture check so they fail cleanly instead of trying to
load a mismatched binary.

## Revised after review

The initial revision of this PR shipped a checked-in PyOgg-derived
`opus.dll` inside a universal wheel. Per maintainer review
(license-metadata drift, universal-wheel-with-arch-specific-DLL
hazard, missing Windows CI coverage, `add_dll_directory` handle
nit), the PR has been rewritten:

| Finding | Resolution |
|---|---|
| **License metadata drift** | New `gateway/LICENSE-THIRD-PARTY` carrying the BSD-3-Clause + Xiph notice. Declared in `gateway/pyproject.toml` `license-files`. README + `gateway/README.md` license sections updated. `SOURCES.md` points to `LICENSE-THIRD-PARTY` as canonical. |
| **Universal wheel ships x86_64 DLL** | `opus.dll` removed from git (added to `gateway/.gitignore`). `publish.yml` gains a `build-windows-wheel` job: bootstraps vcpkg → `vcpkg install opus:x64-windows` → drops DLL into `_libs/` → logs SHA256 → smoke tests → builds wheel → renames `py3-none-any` to `py3-none-win_amd64`. The Ubuntu job produces a clean sdist + `py3-none-any` wheel (no DLL on disk → hatchling artifacts skip it). Publish job downloads both artifact sets. `__init__.py` adds a `platform.machine()` arch gate so even a sdist install on non-x64 Windows skips the DLL search-path setup. |
| **Windows DLL resolution untested in CI** | `build.yml` gains a `gateway-windows-smoke` job. Per PR: bootstraps vcpkg, places `opus.dll`, runs `uv sync --extra tts`, then verifies the bundle is on disk, `_libs/` is on `PATH`, `_dll_dir_handle` is retained, `find_library('opus')` resolves, and an opus encode round-trip succeeds. |
| **`add_dll_directory()` handle discarded** | `__init__.py` retains the returned handle at module scope (`_dll_dir_handle`) so the modern DLL search registration survives garbage collection. The Windows smoke test asserts the handle is non-`None`. |

## Why this matters for users

- **Windows x64** (the vast majority of users): `pip install stackchan-mcp[tts]` just works. The TTS extras import without manual opus install.
- **Windows ARM64 / x86** (rare): `pip install` finds no compatible wheel and either falls back to the sdist with a clean "no opus" failure at import time, or gives a clean "no compatible distribution" error. Either way the user gets actionable feedback instead of an architecture-mismatched DLL crash.
- **Linux / macOS**: unchanged. System `libopus` (apt / brew) provides the runtime as before.

## License compliance

- Gateway Python code: **MIT** (`gateway/LICENSE`).
- Bundled `opus.dll` (win_amd64 wheel only): **BSD 3-Clause + Xiph**
  (`gateway/LICENSE-THIRD-PARTY`).
- Both license files are declared in `pyproject.toml` `license-files`,
  so they ship inside every wheel's `.dist-info/licenses/` directory and
  are visible to license scanners pointed at any distribution form.

## Compatibility

- No public API changes.
- `opuslib` import path on Windows x64 changes from "fails unless user
  installs opus manually" to "works out of the box". On all other
  platforms the import path is unchanged.

## Test plan

- [x] `uv build --wheel` locally produces a `py3-none-any` wheel that
      includes / excludes `opus.dll` based on whether the file is on
      disk at build time (matches the CI Windows / Ubuntu split).
- [x] `gateway/LICENSE-THIRD-PARTY` ships inside the wheel under
      `*.dist-info/licenses/`.
- [x] 251 gateway pytest cases pass with the `__init__.py` changes.
- [x] `windows-latest` smoke test job validated locally on a Windows
      11 x64 host: bundled DLL on disk, PATH prepend, `_dll_dir_handle`
      non-`None`, `find_library('opus')` resolves to the bundled DLL,
      opus encode round-trip OK. (CI runs of this same job will
      surface in PR checks.)

## Notes

- The bundled DLL provenance (CI build from upstream Opus source +
  SHA256 logged per release) replaces the original PyOgg-derived
  binary documented in earlier revisions of `SOURCES.md`. The
  PyOgg-sourced binary is no longer referenced anywhere.
- Local Windows development without CI can mirror the build step:
  `vcpkg install opus:x64-windows` then copy `opus.dll` into
  `gateway/stackchan_mcp/_libs/`. `gateway/.gitignore` keeps the
  local copy out of commits.


kisaragi-mochi (commented):
## Review summary

Thanks for the very thorough SOURCES.md — the license text, provenance
SHA256, and proposed CI-built migration plan all make this PR easy to
review. I ran an adversarial-focused review and there are four findings
to discuss before merge. Two of them I'd consider ship-blockers; one is
a recommendation that probably belongs in this PR scope; one is a small
hardening nit.

### 1. License metadata & public license claims drift (ship-blocker)

The wheel now carries a third-party Opus binary, but
`gateway/pyproject.toml:7-8` still declares only `license = "MIT"` and
ships only `LICENSE` as a license file. README / gateway README still
describe the gateway as MIT-only. `SOURCES.md` includes the BSD-3-Clause
+ Xiph notice (which is great and probably satisfies redistribution),
but downstream users and automated license scanners would reasonably
see the published wheel as MIT-only while it actually contains
BSD-3-Clause/Xiph-licensed binary material.

Suggested resolution:

- Treat the wheel as a composite distribution: include the Opus notice
  in a clearly named `LICENSE-THIRD-PARTY` (or `NOTICES`) file shipped
  inside the wheel/sdist (force-included alongside the DLL).
- Update `gateway/pyproject.toml` `license-files` (or equivalent) so
  the third-party notice is visible to pip's metadata viewers.
- Update README / gateway README license sections: gateway code is MIT,
  but the bundled Windows `opus.dll` is BSD-3-Clause/Xiph with the
  notice carried in the wheel.

### 2. Architecture-specific DLL force-included into a universal wheel (ship-blocker)

`SOURCES.md` documents the bundled binary as x86_64 `win_amd64`, but
`gateway/pyproject.toml:99-100` force-includes it into what is still a
universal wheel — there's no `bdist_wheel --plat-name` setup or per-OS
wheel split in the build pipeline I can see. So `pip install
stackchan-mcp[tts]` on:

- non-Windows (Linux / macOS): pip still resolves the universal wheel,
  which now ships an `opus.dll` that's never loaded but pollutes the
  install footprint.
- Windows ARM64 / Windows x86 (32-bit): pip resolves the same wheel,
  `__init__.py` registers the bundled directory, and `opuslib`
  `find_library('opus')` resolves to a binary that won't load with a
  native-image mismatch — exactly the kind of "looks installed but
  fails at runtime" failure this PR is meant to remove.

This effectively turns a universal wheel into a stealth-Windows-x64
wheel.

Suggested resolution (any of):

- Build/publish a platform-specific `win_amd64` wheel from CI (the
  vcpkg pipeline you mentioned in `SOURCES.md`) and keep non-Windows
  wheels / sdist clean. This was already in your "long-term plan"
  section — promoting it into this PR closes both this finding and
  finding 3 together.
- Or add explicit architecture gating in `__init__.py` (skip the
  `add_dll_directory` setup unless `platform.machine() == 'AMD64'` on
  win32) plus a README note that only x64 Windows Python is supported,
  with the wheel still universal but clearly self-limiting.

The first option scales to other native bundles you may want later
(e.g. a Linux opus binary); the second is the minimal-change version.

### 3. Windows DLL resolution path is untested in CI before publishing (recommended in-PR)

The new behavior runs only on `sys.platform == "win32"`, but
`.github/workflows/build.yml` runs gateway tests on `ubuntu-latest`
only. The publish workflow ships a wheel to PyPI without ever proving
on Windows that:

- the wheel actually contains `_libs/opus.dll`,
- the package import updates `PATH` / `add_dll_directory` as intended,
- `opuslib` resolves and loads the bundled DLL,
- a minimal encode/decode round-trip works against it.

Given that this PR changes pip-distributed native loading for end users
who don't rebuild from source, a CI smoke test in this PR (rather than
a post-merge follow-up) seems warranted. Approximate shape of the job:

```yaml
jobs:
  gateway-windows-smoke:
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with:
          python-version: "3.11"
      - run: pip install ./gateway[tts]
      - run: |
          python -c "import stackchan_mcp, pathlib, sys; \
            assert (pathlib.Path(stackchan_mcp.__file__).parent / '_libs' / 'opus.dll').is_file()"
          python -c "import stackchan_mcp; import opuslib; \
            enc = opuslib.Encoder(48000, 1, opuslib.APPLICATION_VOIP); \
            _ = enc.encode(b'\x00' * 960 * 2, 960); print('opus smoke OK')"
```

For interim signal before the CI job is in place: I ran an equivalent
clean-install + import + encode round-trip on a Windows 11 x64 host
(Python 3.12.10, uv 0.7.17) using the `pr-c-bundle-libopus-windows`
branch. All five smoke checks pass on x64:

```
=== smoke 1: import opuslib (the actual PR #217 fix target) ===
opuslib import: OK

=== smoke 2: _libs/opus.dll bundled into package ===
DLL path: ...\stackchan_mcp\_libs\opus.dll
exists: True
size: 491520

=== smoke 3: PATH prepend confirmed ===
PATH[0]: ...\stackchan_mcp\_libs
matches _libs: True

=== smoke 4: find_library('opus') resolves to bundled DLL ===
find_library opus: ...\stackchan_mcp\_libs\opus.dll

=== smoke 5: Opus encode round-trip (silence frame) ===
encode OK, packet bytes: 57
```

So the user-facing behavior of the patch is verified on
`win_amd64` — that's good news for the import path itself, but it
does not narrow finding 2: I only tested the architecture the bundled
DLL is built for. Non-x64 Windows installs and non-Windows installs
of the same universal wheel remain unverified, which is exactly the
gap finding 2 highlights.

### 4. `os.add_dll_directory()` handle is discarded (nit / hardening)

`stackchan_mcp/__init__.py:46-53` calls
`_os.add_dll_directory(str(_libs_dir))` but does not retain the
returned handle. On Windows that handle is the object used to manage
the registered directory's lifetime; once it's garbage-collected, the
modern resolver registration can become ineffective. The `PATH` update
that runs immediately after will still cover the
`ctypes.util.find_library('opus')` happy path used by `opuslib`, so
the user-visible behavior is fine — but the file comment explicitly
claims the change also supports direct `ctypes.CDLL(...)` and
extension-module dependency resolution, which won't be reliable if
the handle is collected.

Suggested:

```python
_dll_dir_handle = _os.add_dll_directory(str(_libs_dir))  # noqa: F841 — keep alive
```

Or, if you'd rather not introduce a module-global, soften the comment
to "covers the `find_library()` legacy path" only.

### Other notes

- The `[PR #TBD-C]` placeholder in the CHANGELOG entry probably wants
  to become `[PR #217]` (same convention applies to PR #207 which has
  `[PR #TBD-L]`).
- `mergeable: MERGEABLE` — no CHANGELOG rebase needed for this one,
  which is nice given the conflicts on PR #206 / PR #207.

Thanks again for the careful PR body and SOURCES.md.

kisaragi-mochi:
Re-reviewed the revision against the previous review feedback — all four findings from the earlier round are addressed cleanly, especially the `build-windows-wheel` CI job, the `LICENSE-THIRD-PARTY` file, and the `_dll_dir_handle` module-scope retention. Thank you for the thorough revision.

A second-pass adversarial review on the updated branch surfaced five additional findings. The two that affect the release path itself are flagged below as in-PR requests, and the remaining three are carved out as separate follow-up issues to keep the scope of this PR focused on the libopus bundling pipeline it was originally reviewed for.

### In-PR requests (release path safety)

1. **[high] `build-windows-wheel` uses an unpinned vcpkg + the opus.dll SHA256 is only logged, not verified** (`.github/workflows/publish.yml:175-200`). A tag rerun can therefore publish a different native DLL without CI failing, which weakens both supply-chain reproducibility and the license / provenance story for the bundled binary, despite adjacent docs saying the binary is verified against a pinned SHA256. Suggested: pin vcpkg to a reviewed commit / tag, and add an explicit expected SHA256 comparison before the wheel build step (fail the job on mismatch).

2. **[high] The publish workflow smoke test imports `opuslib` without installing the optional `tts` extra** (`.github/workflows/publish.yml:207-220`). `uv sync --frozen` does not install `opuslib` because it is declared under the optional extras. The immediate smoke test then imports `opuslib`, so the tag release path can fail with `ModuleNotFoundError` while PR CI continues to pass (because `build.yml` uses `--extra tts`). This is a release-only failure mode in the exact job that creates the Windows wheel. Suggested: change to `uv sync --frozen --extra tts` (or equivalent) in `build-windows-wheel`.

### Carved out as follow-up issues

The remaining three findings are tracked for separate follow-up so this PR's scope stays tight:

- #221 — gateway: internal WHEEL Tag inconsistency in win_amd64 retag (the `.dist-info/WHEEL` `Tag:` stays `py3-none-any` after filename retag; `pip` resolves and installs cleanly today, but the metadata mismatch can produce inconsistent license-scanner reports and may be rejected by stricter tooling)
- #222 — gateway: polish primary LICENSE + Xiph patent grant verbatim (primary `LICENSE` does not yet reference the bundled native binary boundary; the Xiph grant text is named in `LICENSE-THIRD-PARTY` but linked via IETF IPR URL rather than reproduced)

The current artifact installs and works on Windows 11 x64 today (your smoke test demonstrates this), so neither carved-out item blocks merge.

Thanks again for the careful revision.

kisaragi-mochi:
Rebased and force-pushed to resolve the `CHANGELOG.md` conflict caused on our side by the recent merge of #212 (which landed the `send_pcm_audio` helper bullet in the same `[Unreleased] / Gateway` block as your libopus bundle bullet). The rebase only re-anchored your 5 commits on top of the new `main` HEAD and merged the `[Unreleased]` section so the two bullets sit side-by-side; bullet contents, code diff, and commit authorship are unchanged. No action needed on your end.

Codex review is being kicked off against the rebased HEAD; will update labels (`needs-review` → `ready-to-merge`) once that comes back clean.

kisaragi-mochi:
The Windows wheel SHA256 has been pinned to the actual value from a `workflow_dispatch` run on a temporary branch:

```
EXPECTED_OPUS_DLL_SHA256: 99cd252b09891398cefab997be8cdb1cebc9a40cafa1382c88ad32069560a864
```

Source: vcpkg pinned tag `2026.04.27` (Opus 1.5.2 port-version 1) built on `windows-latest` runner, captured from the `Verify opus.dll SHA256` step output. The temp branch on this repo has been deleted now that the hash is in.

Codex re-review is being kicked off against the updated HEAD. Once that comes back clean, the label will move to `ready-to-merge`.

kisaragi-mochi:
Codex re-review on the rebased HEAD (post hash pin): 1 P2 finding, which is a known carve-out — #221 (internal WHEEL `Tag:` metadata stays `py3-none-any` after the filename rename to `py3-none-win_amd64`).

#221 was already opened from the second-pass adversarial review of this PR with the explicit scope decision that the wheel-build mechanism polish belongs in a follow-up, since the current artifact ships and works for end users today (Windows 11 x64 smoke test passes; `pip install` resolves and runs cleanly).

The license-polish carve-outs (#222) are similarly tracked separately.

Moving the label to `ready-to-merge`. No further changes needed on this PR — review-side is closed out.

Issue #95, opened 2026-05-12
feat(firmware): bump CoreS3 mic input_gain from 30 to 37.5 dB


kisaragi-mochi:
## Context

PR #93 (gateway-side `listen()`) end-to-end verification on a real CoreS3 surfaced that distant speech (~30 cm from the device mic) returns an empty transcription even though the wire path (Opus frames, decode, transcribe) is working correctly. Close-range speech (5 cm or less) transcribes reliably.

See https://github.com/kisaragi-mochi/stackchan-mcp/pull/93#issuecomment-4430347622 for the trial table.

## Hypothesis

The CoreS3 `input_gain_` is currently `30` dB in `firmware/main/boards/m5stack-core-s3/cores3_audio_codec.cc:17`. Two other boards in this firmware explicitly raise theirs to `37.5` dB with comments documenting the same problem:

- `firmware/main/boards/esp-box-lite/box_audio_codec_lite.cc:195`: `esp_codec_dev_set_in_gain(input_dev_, 37.5);`
- `firmware/main/boards/df-k10/k10_audio_codec.cc:180`: `esp_codec_dev_set_in_gain(input_dev_, 37.5); //麦克风增益解决收音太小的问题`

The Chinese comment on `df-k10` translates to "mic gain to solve the too-quiet recording problem", which is the same symptom observed on CoreS3.

37.5 dB is +7.5 dB over the current setting, roughly 2.4x linear gain.

## Proposed change

Change `firmware/main/boards/m5stack-core-s3/cores3_audio_codec.cc:17` from `input_gain_ = 30;` to `input_gain_ = 37.5;`. Single-line firmware change, no schema/protocol impact, no breaking change.

## Risks

- Slight increase in background noise pickup. PR #93 trial 4 confirmed that pure silence still returns an empty transcription with `faster-whisper` `base`, so a modest gain bump is not expected to flood the recogniser with hallucinations.
- Speaker feedback into mic is mitigated by the existing `input_reference_` echo-cancellation path; this change does not touch that wiring.

## Validation

Re-run the PR #93 distant-speech trial (Trial 1) after the change. Acceptance: distant speech (~30 cm) returns a non-empty transcription comparable to the close-range trials.

## Out of scope

- Per-call runtime gain control via the `listen()` MCP tool. This Issue is scoped to the firmware-side default.
- Other boards. The 37.5 dB value is already in use on `esp-box-lite` / `df-k10`; no change to those.

Refs #91, #93, #8.

PR #225, opened 2026-05-22
fix(firmware): keep idle after tts.stop, remove auto-listening recovery

maha0525:
## Summary

When the stack-chan TTS pipeline stalls, the deferred `tts.stop` event currently lands long after the user expected the conversation to be quiescent, and the device drops into Listening mode for up to 30 s. The gateway picks up the ambient room audio and posts it as a user utterance. This PR removes the auto-listening fallback after `tts.stop` so the device returns to Idle and the listening boundary is set explicitly by whoever wants to continue the conversation.

## Why

`stackchan-mcp` is an MCP gateway, not a standalone xiaozhi-style conversational agent. Listening must be triggered explicitly — either by the user (touch / button / external command) or by the AI (gateway-issued `StartListening`). Loop-style `Listening → Speaking → Listening` flows belong on the gateway side, where the gateway can reason about the conversation's actual state.

The upstream xiaozhi behaviour of auto-entering Listening after every TTS utterance is incompatible with this model. Concretely:

- In a long playback I observed `task_wdt` timeouts on CPU 0 (`audio_input`) firing every 10 s for over 2 minutes. When the pipeline finally recovered the deferred `tts.stop` event arrived, and the existing `listening_mode_ != kListeningModeManualStop` branch sent the device into Listening with a 30 s timeout.
- `GetDefaultListeningMode()` returns `kListeningModeAutoStop` (no-AEC) or `kListeningModeRealtime` (AEC on) — never `kListeningModeManualStop`. So in the gateway-driven path, the `ManualStop` guard is effectively dead and the auto-listening branch always runs.
- The 30 s of recorded ambient room audio gets posted to the gateway as a user utterance, multiple times per session, with predictably confusing downstream effects.

The existing comment around `HandleStartListeningEvent()` already calls out that "Manual-stop is also the right default for gateway-driven capture: the gateway controls the exact stop boundary" — this PR brings the `tts.stop` path in line with that intent.

## Change

`application.cc`, the `Schedule` callback for `tts.stop`:
- Before: branch on `listening_mode_ == kListeningModeManualStop` → `Idle`, else `Listening`.
- After: unconditional `SetDeviceState(kDeviceStateIdle)` when leaving `Speaking`.

The `OnTtsStop()` board callback is unchanged (still fires unconditionally so the avatar mouth animation stops even when an external abort moved the device out of Speaking before `tts.stop` arrived).

## Test plan

- [x] Built and flashed on a CoreS3 + Si12T stack-chan unit (ESP-IDF 5.5.4) and observed `StateMachine: State: speaking -> idle` on every `tts.stop` for several long playbacks.
- [x] Confirmed no further runaway ~30 s ambient-audio captures after the fix (over a multi-hour session that previously produced 3+ such events per hour).
- [ ] Real-device acceptance from someone running a standalone xiaozhi-style UX (not me — I only run the MCP gateway path; would appreciate a sanity check from anyone who relies on auto-listening for that use case).

## Notes for review

Happy to gate this behind a `CONFIG_*` opt-in if the auto-listening behaviour is load-bearing for non-gateway users; let me know which direction you'd prefer. My read is that the gateway-driven path is the more common one for stack-chan deployments, but you have a much better sample of users than I do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,28 @@ documented-only.
 
 ### Firmware
 
+- Changed: the device no longer auto-enters Listening after a TTS
+  utterance ends. The `Application::OnIncomingJson` handler for the
+  `tts.stop` event used to fall through to
+  `SetDeviceState(kDeviceStateListening)` whenever `listening_mode_`
+  was anything other than `kListeningModeManualStop`, mirroring the
+  upstream xiaozhi conversational-agent UX. In an MCP-gateway
+  deployment that auto-listening path is a footgun: when the TTS
+  pipeline stalls (we observed `audio_input` task watchdog timeouts
+  firing every 10 s for over two minutes during long playback), the
+  deferred `tts.stop` event arrives long after the user expected the
+  conversation to be quiescent, and the device then records ~30 s of
+  ambient room audio that the gateway happily posts as a user
+  utterance — repeatedly per session. With this change `tts.stop`
+  unconditionally returns the device to Idle; listening is now started
+  only by the user (touch / button / external command) or by the AI
+  (gateway-issued `StartListening`). Loop-style `Listening → Speaking
+  → Listening` flows belong on the gateway side, where the gateway
+  can reason about the conversation's actual state. Avatar mouth
+  animation cleanup via `OnTtsStop()` still fires unconditionally.
+  Contributed via
+  [PR #225](https://github.com/kisaragi-mochi/stackchan-mcp/pull/225).
+
 - Added: `self.port_b.ws2812.{init, set_pixel, set_strip, refresh, clear}`
   MCP tools — five generic tools to drive any WS2812-compatible LED
   strip attached to the official kit's Port B (CoreS3 HY2.0-4P digital
--- a/firmware/main/application.cc
+++ b/firmware/main/application.cc
@@ -529,11 +529,28 @@ void Application::InitializeProtocol() {
             } else if (strcmp(state->valuestring, "stop") == 0) {
                 Schedule([this, &board]() {
                     if (GetDeviceState() == kDeviceStateSpeaking) {
-                        if (listening_mode_ == kListeningModeManualStop) {
-                            SetDeviceState(kDeviceStateIdle);
-                        } else {
-                            SetDeviceState(kDeviceStateListening);
-                        }
+                        // stackchan-mcp is an MCP gateway, not a standalone
+                        // xiaozhi-style conversational agent. Listening must
+                        // be triggered explicitly — either by the user
+                        // (touch / button / external command) or by the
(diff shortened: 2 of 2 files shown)


kisaragi-mochi:
Codex review against main: 0 findings.

The dead-code observation on the existing `kListeningModeManualStop` branch holds: `GetDefaultListeningMode()` returns only `kAutoStop` or `kRealtime`, so the prior `Idle` branch was unreachable in practice. The unconditional `Idle` transition here also aligns with the existing rationale in `HandleStartListeningEvent()` ("Manual-stop is also the right default for gateway-driven capture") — the two `tts.stop` and `StartListening` paths now converge on the same intent.

On the design question about a `CONFIG_*` opt-in for auto-listening: the gateway-driven model is what this repository targets, and the standalone xiaozhi-style auto-listening loop is out of scope here. The unconditional `Idle` transition is the preferred direction; we'd rather not reintroduce the footgun behind a flag.

Will move the label to `ready-to-merge` after a real-device sanity flash on this end. Planning to amortise the flash cycle across the upcoming firmware PRs (#210, #211).

kisaragi-mochi:
Real-device verification on a CoreS3 + SCS0009 + Si12T stack-chan unit (ESP-IDF 5.5.4 build via the canonical `python ./scripts/release.py stackchan` Docker path) confirms this PR.

### Step 1 — avatar / boot / WS

- `get_status` after flash returns `connected: true, initialized: true, tools_count: 30` (= matches the post-#223 baseline; no missing tools).
- `set_avatar(happy)` brings up the avatar on the LCD; the avatar pipeline and WS dispatch are intact.

### Step 2 — PR core scope

- `say()` with a ~10 s voicevox utterance (`duration_ms: 10320, frame_count: 172`) completes cleanly.
- Immediately after `tts.stop`, the device transitions straight to Idle and **does not enter Listening**. Held silent for 30+ seconds afterwards with no further audio frames posted to the gateway — i.e. the runaway ~30 s ambient capture that previously fired on every `tts.stop` is gone.

### Step 3 — regression check on manual listen

- `listen(duration_ms=8000, motion="face-only")` triggers the listening mode transition correctly (`frame_count: 129, duration_ms: 7740`); the thinking avatar shows during the capture window and the `play_popup_on_listening_` cue (PR #207) fires as expected.
- STT transcription accuracy is **out of scope** for this PR — the current 30 dB mic input_gain is the limiting factor, which is already tracked separately under Issue #95 (`bump CoreS3 mic input_gain from 30 to 37.5 dB`). At the wire level, the manual listen path is unaffected by this change, so no regression.

### Conclusion

The fix removes the auto-listening fallback after `tts.stop` exactly as intended for the MCP-gateway-driven path, with no observable regression on manual listen. Moving the label to `ready-to-merge`.

PR #240, opened 2026-05-29
fix(firmware/ws): add active keepalive to recover from silent network breaks (#239)

### LINKED ISSUE #239: bug(firmware): no active WS keepalive — silent network break leaves device in stuck-connected state
## Symptom

On 2026-05-29 ~00:32:25, the WebSocket connection between device and gateway broke silently. The gateway detected the disconnect via the websockets-library default `ping_timeout` (20 s) and logged:

```
00:32:25,608 [INFO] stackchan_mcp.esp32_client: ESP32 disconnected: device=44:1b:f6:df:58:e0
```

The device, however, did **not** detect the disconnect at all — no `WS: Websocket disconnected`, no `EspTcp: TCP receive failed`, no WiFi event. It remained in a "connected" state with zero reconnect attempts until manually rebooted ~19 hours later.

## Repro context

Device on LAN WiFi to gateway (`192.168.0.9:18765`). Around the failure window, a brief device + gateway-side network outage occurred. Verified from Windows Event Log on the gateway PC:

```
0:32:25 [Warning]  Dhcp-Client: DHCP address renewal failed
0:34:33 [Error]    Dhcp-Client: No address assigned
0:40:12 [Info]     WLAN-AutoConfig: disconnected from network
0:40:14 [Info]     WLAN-AutoConfig: reconnected successfully
```

The `0:32:25` DHCP renewal failure matches the gateway's `ESP32 disconnected` log timestamp to the second. The PC's WiFi reconnected at `0:40:14`, but the device remained in its stuck-connected state with no recovery attempt afterwards.

Earlier on the same day, normal SAIVerse restarts (22:32, 23:18, 23:50) all produced clean disconnect→reconnect cycles on the device side. The 00:32 break differs only in that no detection signal reached the device — neither the gateway's WS close (after `ping_timeout`) nor a TCP FIN/RST could traverse the broken path.

The trigger is not reproducible on demand, but the failure mode is deterministic given the trigger.

## Cause analysis

The device-side WS layer (managed_component `78__esp-ml307`) uses purely passive disconnect detection — it relies on a TCP FIN/RST or WS close frame from the server arriving at `recv()`. This works for normal disconnects but assumes the close packet can traverse the network back to the device.

In the 00:32 case the underlying path was broken in both directions, so the gateway's close (after its own `ping_timeout`) was sent into a black hole.

Device-side, there is no fallback:

- `WebSocket::Ping()` exists in the library (`esp-ml307/src/web_socket.cc:253`) but is never called from firmware (grep zero in `firmware/main/`)
- `recv()` has no application-level timeout (`esp-ml307/src/esp/esp_tcp.cc:135`)
- No `SO_KEEPALIVE` / `SO_RCVTIMEO` set on the socket

So when the server stops responding, the device has no way to notice.

Passive disconnect detection via server-side close is a valid WS client pattern in itself; the gap is specifically silent-path-break coverage.

## Suggested direction

Active firmware-side ping with a per-side timeout would let the device detect dead connections independently of server config. This is implementable in `websocket_protocol.cc` using the existing `WebSocket::Ping()` API — no library changes required.

Library-level alternatives (`SO_KEEPALIVE`, recv timeout) would belong upstream in `esp-ml307`; happy to mirror there if preferred.

Willing to put up a firmware-side PR if the direction is acceptable.

## Related code

- `gateway/stackchan_mcp/esp32_client.py:469` — `websockets.serve()` invocation (default ping params)
- `gateway/stackchan_mcp/esp32_client.py:675-676` — `ConnectionClosed` catch + disconnect log
- `firmware/managed_components/78__esp-ml307/src/web_socket.cc:253` — `WebSocket::Ping()` (defined but unused)
- `firmware/managed_components/78__esp-ml307/src/esp/esp_tcp.cc:135` — `recv()` with no timeout

maha0525:
Closes #239.

Adds an active firmware-side WebSocket keepalive so the device can detect and recover from silent network breaks — the failure mode where a brief WiFi outage drops packets in both directions before any TCP FIN/RST can propagate, leaving the device stuck in a "connected" state with no reconnect attempt.

## Mechanism

A periodic timer (every 15 s):

- Checks `now - last_received_us_` against a 60 s dead-threshold. If exceeded → `websocket_.reset()` to trigger the existing `OnDisconnected` → `ScheduleReconnect` path.
- Otherwise sends a WS Ping to actively probe the path.

`last_received_us_` is updated in the existing `OnData` callback, so any text/binary frame from the gateway (including normal MCP polls) refreshes the liveness signal.

The timer lifecycle mirrors `reconnect_timer_`:

- Created once in the constructor, deleted in the destructor.
- Started after a successful handshake in `OpenAudioChannelInternal()`.
- Stopped from `OnDisconnected` and the destructor.
- Callback hops to the main task via `Application::Schedule` to keep `websocket_` touches off `ESP_TIMER_TASK`.

## Verification on real hardware (ESP32-S3 on WiFi)

1. **Regression check** — Normal connect, SAIVerse restart cycles, and avatar fetch unaffected. `Keepalive started (check every 15 s, dead threshold 60 s)` appears right after every successful Session ID.

2. **Silent break detection** — Simulated by a Windows firewall `Block`-action rule on port 18765 (DROPs SYN/data rather than RSTing them). Observed log:

   ```
   23:46:58.345  WS: Session ID: a4533552-...
   23:46:58.345  WS: Keepalive started (check every 15 s, dead threshold 60 s)
   ( ... firewall rule added ... )
   23:49:43.348  WS: Keepalive timeout: no data frame received in 72864 ms (>= 60 s threshold) — forcing reconnect
   23:49:43.348  WS: Websocket disconnected
   23:49:43.370  WS: Schedule websocket reconnect in 5 seconds
   ```

   Reconnect attempts looped as expected. Once the firewall rule was removed, the next reconnect succeeded and `Keepalive started` re-appeared.

## Tuning notes

- `WEBSOCKET_KEEPALIVE_INTERVAL_MS = 15 s` and `WEBSOCKET_KEEPALIVE_DEAD_TIMEOUT_MS = 60 s` are sized to comfortably cover the gateway's normal ~30 s MCP poll cadence (`avatar_loader`) plus margin so a single missed poll doesn't trip the watchdog.
- Server-side ping_interval is not negotiated, so the firmware tracks liveness via its own data-frame timestamp rather than relying on any gateway-side cadence assumption.

## Notes

- Pure firmware-side change; no `esp-ml307` library modifications required (`WebSocket::Ping()` is already exposed).
- Format strings use `%u` rather than `%llu` to stay compatible with ESP-IDF's nano-printf.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,13 @@ documented-only.
 
 ### Firmware
 
+- Added: active firmware-side WebSocket keepalive that detects silent
+  network breaks and triggers the existing reconnect path. A periodic
+  Ping (every 15 s) probes the connection; the Pong response refreshes
+  a liveness timestamp via the new `WebSocket::OnPong` callback
+  (esp-ml307 #49). If no frame of any kind arrives within 60 s, the
+  connection is considered dead and a graceful reconnect is forced —
+  no device reboot required. Closes #239.
 - Changed: the device no longer auto-enters Listening after a TTS
   utterance ends. The `Application::OnIncomingJson` handler for the
   `tts.stop` event used to fall through to
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -79,6 +79,30 @@ WebsocketProtocol::WebsocketProtocol() {
         ESP_LOGE(TAG, "Failed to create reconnect timer; auto reconnect will not be available");
         reconnect_timer_ = nullptr;
     }
+
+    // Keepalive timer: periodic check that drives both active WS ping
+    // emission and silent-path-break detection. See issue #239 and the
+    // header comment on keepalive_timer_ for the full rationale. The
+    // callback hops to the main task via Application::Schedule to mirror
+    // the reconnect timer pattern (avoid touching websocket_ from
+    // ESP_TIMER_TASK).
+    esp_timer_create_args_t keepalive_timer_args = {
+        .callback = [](void* arg) {
+            auto protocol = static_cast<WebsocketProtocol*>(arg);
+            auto alive = protocol->alive_;
+            Application::GetInstance().Schedule([protocol, alive]() {
+                if (!alive->load()) {
+                    return;
+                }
+                protocol->OnKeepaliveTick();
+            });
+        },
+        .arg = this,
+    };
+    if (esp_timer_create(&keepalive_timer_args, &keepalive_timer_) != ESP_OK) {
+        ESP_LOGE(TAG, "Failed to create keepalive timer; silent-break recovery will not be available");
+        keepalive_timer_ = nullptr;
+    }
 }
 
 WebsocketProtocol::~WebsocketProtocol() {
@@ -92,6 +116,11 @@ WebsocketProtocol::~WebsocketProtocol() {
     if (reconnect_timer_ != nullptr) {
         esp_timer_delete(reconnect_timer_);
         reconnect_timer_ = nullptr;
+    }
+    StopKeepaliveTimer();
+    if (keepalive_timer_ != nullptr) {
+        esp_timer_delete(keepalive_timer_);
+        keepalive_timer_ = nullptr;
     }
     websocket_.reset();
     if (event_group_handle_ != nullptr) {
@@ -388,6 +417,11 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_aud
         websocket_->SetHeader("Client-Id", Board::GetInstance().GetUuid().c_str());
 
         websocket_->OnData([this, notify_disconnect, arm_audio_channel](const char* data, size_t len, bool binary) {
+            // Any data frame from the gateway counts as proof-of-life for
+            // the keepalive watchdog (see issue #239). Update before any
+            // early-return below so dropped/late frames still refresh the
+            // liveness signal.
+            last_received_us_.store(esp_timer_get_time(), std::memory_order_release);
             if (binary) {
                 // Drop binary frames before parsing when the device is not
                 // speaking. This mirrors Application::OnIncomingAudio's
@@ -468,6 +502,11 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_aud
         websocket_->OnDisconnected([this, notify_disconnect, disconnected_after_hello]() {
             audio_channel_open_.store(false);
             transport_connected_.store(false);
+            // Stop the keepalive watchdog whenever the transport drops —
+            // whether the close was server-initiated, keepalive-forced,
+            // or intentional. StartKeepaliveTimer() will re-arm it on
+            // the next successful handshake. See issue #239.
+            StopKeepaliveTimer();
             // notify_disconnect carries this socket's reconnect intent.
             // ParseServerHello() arms it (true) once the handshake
             // completes; intentional teardown paths (CloseAudioChannel,
@@ -583,6 +622,11 @@ bool WebsocketProtocol::OpenAudioChannelInternal(bool report_error, bool arm_aud
         transport_connected_.store(true);
         reconnect_interval_ms_ = WEBSOCKET_RECONNECT_INITIAL_INTERVAL_MS;
         StopReconnectTimer();
+        // Seed the keepalive baseline with "now" so the dead-timeout
+        // window starts from the connect time, then arm the periodic
+        // check. See issue #239.
+        last_received_us_.store(esp_timer_get_time(), std::memory_order_release);
+        StartKeepaliveTimer();
 
         if (on_connected_ != nullptr) {
             on_connected_();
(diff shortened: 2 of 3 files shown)


kisaragi-mochi:
Thank you for tackling #239 — the silent network break failure mode is exactly the kind of edge case that's hard to even file an issue against without the kind of detailed Windows Event Log + device-side log correlation you provided. The active firmware-side keepalive direction is sound, and on real hardware your verification covered the explicit silent-break path well.

I dug into the design once more on this end before merging, and want to share what I found and a couple of follow-up requests around it.

## What I checked

The PR description grounds the 60s `WEBSOCKET_KEEPALIVE_DEAD_TIMEOUT_MS` threshold in "the gateway's normal ~30 s MCP poll cadence (`avatar_loader`) plus margin." I went looking for that cadence on the stackchan-mcp side to confirm the margin holds, and the picture I came back with is a bit different from what the description assumes:

- The string `avatar_loader` doesn't appear anywhere in this repo (gateway/, firmware/). The closest match is the `load_avatar_set` MCP tool + the `avatar_set_fetch` WebSocket flow added in #210 / #211, but those are explicitly one-shot (on-demand) call paths, not periodic polls.
- `gateway/stackchan_mcp/` has no `asyncio.sleep`-based periodic loop that talks to the device. MCP traffic to the device is generated when an LLM agent calls a tool, and the stdio MCP layer is fully passive otherwise.
- The `avatar_loader.py` that *does* exist lives in `maha0525/saiverse-stackchan-addon`, and from reading the source it's event-driven (fires once when a persona enters a Vessel Building, skips the load if the same persona stays). It's not a 30s periodic poll either.
- The `saiverse-stackchan-addon` README mentions "heartbeat + identity-aware unregister" for TCP half-open coverage, but that appears to sit at the SAIVerse server / addon layer, not at the firmware ↔ gateway layer this PR is operating on. Whether the addon-layer heartbeat actually causes data frames to arrive at the firmware's `OnData` callback (and thus refresh `last_received_us_`) isn't obvious from reading the addon source alone.

Putting it together, on a bare stackchan-mcp deployment (no SAIVerse, no addon) there's nothing that would generate periodic LLM-independent traffic to the device. An idle window of >= 60s is a normal state for that deployment — the user is thinking, the agent finished speaking and is waiting for the next turn, the user has stepped away briefly — and in that state the keepalive would currently trigger `websocket_.reset()` every minute, kicking off the existing reconnect path. That's the part I wanted to check with you before merging.

## A note on the PR scope

Looking back at the PR body, I don't see an explicit treatment of the SAIVerse-absent case — the `Verification` block and the `Tuning notes` are both written from a SAIVerse-paired perspective (avatar_loader cadence, SAIVerse restart cycles), and `Tuning notes` actually grounds the 60s threshold in that environment.

I wanted to check whether the bare-stackchan-mcp deployment was considered at design time and we just missed seeing it documented, or whether the 60s threshold was sized specifically for your typical environment and the no-addon case wasn't in scope. Either is totally reasonable as the starting point — just helps us know which direction to take from here.

## Request A (primary)

Would you be able to add a check on your end for the steady-idle case? Start a session, do nothing for a couple of minutes, and confirm that the device log doesn't show recurring `Keepalive timeout` / forced-reconnect lines.

If your deployment turns out to hit that pattern too, then whatever fix lands on your side — likely an application-level keepalive somewhere in the addon / SAIVerse layer that generates a real data frame to the device — would also cover the bare stackchan-mcp deployment, so a single fix lands cleanly across both environments.

## Request B (conditional, only if A finds no recurring reconnects in your environment)

If A turns out clean — meaning some part of the SAIVerse / addon stack *is* producing periodic traffic that reaches the firmware's `OnData` and we just missed it on our reading — then on a bare stackchan-mcp deployment that traffic isn't there, and the same >= 60s idle window will trip the keepalive. In that case I'd ask for one of:

- **B1**: Adding an application-level keepalive on the stackchan-mcp side (gateway → device, periodic enough to refresh `last_received_us_`) so the firmware keepalive becomes sufficient on its own without depending on a downstream component, **or**
- **B2**: Making the firmware keepalive opt-in (e.g. `CONFIG_STACKCHAN_WS_KEEPALIVE` defaulting to `n`, with the addon's `sdkconfig.defaults.local` flipping it on). The C++ logic itself stays as-is, just gated.

Either keeps the SAIVerse + addon deployment with its silent-break recovery while not changing the default behavior for users running stackchan-mcp on its own.

## Where this is coming from

`avatar_loader` being a SAIVerse-side concept means we want to be careful about absorbing SAIVerse-environment-specific behavior into the stackchan-mcp base. The collaboration on the PCM streaming series has been great and we want to keep merging compatible contributions — but for behavior that's specifically caused by a SAIVerse component not being present in the bare deployment, that's not really our scope to cover from this side. Hopefully A makes it a non-issue: if everyone sees the reset loop, the fix probably lives downstream of where this PR is operating.

Also a small one while we're here — the `CHANGELOG check` workflow is failing because there's no entry under `[Unreleased]` > `### Firmware`. Could you add a short line there (e.g. `Added: active firmware-side WebSocket keepalive that detects silent network breaks and triggers the existing reconnect path (#239).`) in the next push?

No rush on any of this — happy to keep the PR open while we figure A out.

maha0525:
Thanks again — and your read was right on the money. Digging into Request A turned the whole thing on its head, so here's where it landed.

## Request A (steady-idle test)

On my SAIVerse-paired setup, idle is clean — no recurring `Keepalive timeout` / forced reconnects over multi-minute idles. But you were right that this isn't representative of a bare deployment: the only thing keeping `last_received_us_` fresh during idle was the SAIVerse addon's session-reconcile poll (a `get_device_info` round-trip every 30 s). I'd mis-attributed that cadence to `avatar_loader` in the PR body — the avatar fetch is event-driven, exactly as you found; the 30 s cadence is the reconcile loop. On a bare `stackchan-mcp` deployment that poll doesn't exist, so a >60 s idle window would indeed trip the watchdog and reconnect-loop. Precisely the scope concern you raised.

## Root cause, one level down

The deeper problem is that this PR's liveness signal is "an application *data* frame arrived," which has no guaranteed source. The standard WebSocket mechanism for this is Ping/Pong: the firmware already sends a Ping every 15 s, but `esp-ml307`'s `WebSocket` silently discards the incoming Pong (the `0xA` branch was a bare `break;`). So the firmware couldn't use its own Ping/Pong as the liveness signal and fell back to data-frame timestamps — which is what created the implicit dependency on application traffic.

## Fix

I opened a small upstream PR on esp-ml307 to surface the Pong: 78/esp-ml307#49 (adds an `OnPong` callback mirroring `OnData`/`OnError`; no-op if unused). With it, the firmware keepalive becomes self-contained — it refreshes liveness from the Pong answering its own Ping, with zero dependence on gateway or application traffic. That makes it correct for a bare `stackchan-mcp` deployment *and* the SAIVerse one, with no gateway-side periodic loop — so neither Request B1 nor the opt-in B2 is needed.

## Verified on real hardware (ESP32-S3, WiFi)

- Idle (3.5 min): no false reconnects.
- Pong reaches the callback every ~15 s ping cycle on an idle-but-healthy link (measured) — liveness stays fresh with no application traffic at all.
- Silent break (firewall dropping the WS port): detected at ~75 s and recovered via a **graceful WebSocket reconnect — no device reboot** (uptime counter continuous across the reconnect).

## Plan for this PR

Once esp-ml307#49 lands and is released, I'll revise this branch to swap the data-frame-timestamp approach for the Pong-based one, bump the esp-ml307 dependency, and add the `[Unreleased] > Firmware` CHANGELOG entry. I'll hold it as-is until then. Thanks for absorbing the rebase in the meantime — and for pushing back on the scope, it led to a cleaner fix.

kisaragi-mochi:
Acknowledged — thank you for the deep root-cause analysis here, and especially for following up with the upstream patch at https://github.com/78/esp-ml307/pull/49 (merged 2026-06-03). Surfacing the Pong via `OnPong` and rebasing this PR on top of that lands a self-contained liveness check, which is the right shape — it removes the implicit dependency on application traffic and makes the same fix correct for both bare `stackchan-mcp` and SAIVerse-paired deployments.

Confirmed status:

- `78/esp-ml307#49` is merged on the upstream `main` branch.
- The ESP-IDF Component Registry currently publishes `78/esp-ml307@3.6.5` (created 2026-03-07); a release including the `OnPong` callback is pending an upstream registry push.
- `stackchan-mcp` firmware pins `78/esp-ml307: ~3.6.5`, so a `3.6.x` / `3.7.x` release will be picked up automatically once available.

No rush from this side. Happy to keep this PR open with `blocked` while we wait for the upstream release; rebase + `[Unreleased] > Firmware` CHANGELOG entry whenever it lands. Thanks again for taking it upstream rather than working around it locally — that is the better long-term shape for this codebase.

kisaragi-mochi:
Update: 78/esp-ml307 version 3.6.6 was published to the ESP Component Registry on 2026-07-15, and it includes the OnPong callback (upstream commit 52b49eb2, merged 2026-06-03 via PR #49 on that repo). This satisfies the existing tilde-3.6.5 pin in firmware/main/idf_component.yml, so the rebase described in the plan above should be unblockable now.

Flagging this proactively as part of a sweep for anything waiting on an external event that may have quietly already happened, so it doesn't sit longer than necessary once the release it was waiting on has landed.

kisaragi-mochi:
Looked over the current head after the OnPong rebase — the keepalive path still looks right. The remaining conflicts were on our side: `main` moved a long way (including #361) while this sat.

Merged `origin/main` into `fix/ws-keepalive-silent-break` as cea04d5 and pushed it here. Kept both the OnPong keepalive members and the reconnect-timer guard that landed on main in the meantime. The `[Unreleased] > Firmware` CHANGELOG entry for #239 is on top of current main.

CI is green and the PR is mergeable. Nothing needed on your end — thanks for waiting this out, and for taking the liveness check upstream via 78/esp-ml307#49.

Issue #134, opened 2026-05-15
feat(firmware): boot-time servo climb feels stepped / non-smooth despite 4s duration (PR #125 follow-up)


kisaragi-mochi:
## Summary

After PR #125 lengthened `BOOT_INIT_MOVE_MS` from 1000 ms to 4000 ms (`~45°/s` → `~11°/s`), the boot-time climb feels at the right *speed*, but the motion itself appears stepped / frame-by-frame rather than smooth. The slower duration makes a pre-existing quantization that was invisible at the faster `1s` setting now visible to the user.

This is a follow-up to PR #125; the duration change is correct and should not be reverted. The improvement target is the *smoothness* of the interpolation along that 4 s curve, separately tracked from the user-driven `move_head` work in #129 (which deliberately keeps `BOOT_INIT_MOVE_MS` out of its scope).

## Observed behavior

Verified on a clean `firmware-v1.4.1` install (`merged-binary.bin` at `0x0` + on-device captive-portal setup), with the pre-boot pose set to `{yaw=0, pitch=10}` via `move_head` and then hard-reset:

```
Boot pre-init ReadPos: yaw_raw=461 pitch_raw=652 tick=133
Restored pitch_motion_.current_deg=10 from ReadPos=652 (clamped to safe range 0..88)
... 4.1 s later ...
Boot-time servo init complete: target yaw=0 pitch=45 (move=4000ms),
  post-ReadPos: yaw_raw=461 pitch_raw=762, elapsed_ms=4100
```

The head climbs from `pitch=10°` to `pitch=45°` over 4.1 s, but the climb is visually segmented rather than continuous — distinct "tick" jumps are visible to the eye on the CoreS3 + SCS0009 hardware.

## Hypothesized root cause

Three concurrent contributors interact in `firmware/main/boards/stackchan/stackchan.cc`:

1. **50 Hz tick + linear interpolation**: `ServoTaskMain()` runs every `MOTION_TICK_MS = 20 ms` and computes the next per-axis angle via linear interpolation between current and target. Over a `35°` move spread across 200 ticks, each tick advances ~`0.175°`.

[...]

Issue #129, opened 2026-05-15
feat(firmware): tune user-driven StackChan servo motion responsiveness


kisaragi-mochi:
## Background

Issue #88 surveyed the current StackChan servo motion path and the relevant official/community implementations before changing firmware behaviour. The main outcome is that the first implementation step should focus on user-driven `move_head` / `self.robot.set_head_angles` responsiveness, not boot-time servo positioning.

Boot-time motion is already handled separately:

- #115 / PR #117 added boot-time neutral positioning (`yaw=0`, `pitch=45`).
- #121 / PR #125 slowed the boot-init climb from 1 s to 4 s.
- #100 remains the hang/recovery methodology issue. Its latest evidence says `{"yaw":-144,"pitch":-194}` alone is not sufficient hang proof after PR #124.

## Current implementation to tune

`firmware/main/boards/stackchan/stackchan.cc`:

- `WriteHeadAngles(yaw, pitch, duration_ms = MOTION_DEFAULT_DURATION_MS)` updates motion targets only.
- `ServoTaskMain()` runs every `MOTION_TICK_MS = 20` ms.
- Each tick linearly interpolates current angle toward the target.
- Each moving axis is sent as a separate `WritePos()` call:
  - yaw first
  - `kInterFrameGap = 10 ms`
  - pitch second
- Each per-tick command uses `MOTION_PER_WRITE_TIME_MS = 30` and `speed = 0`.
- User-driven calls currently use `MOTION_DEFAULT_DURATION_MS = 600`.
- `BOOT_INIT_MOVE_MS = 4000` is separate and should not be reduced in this issue.

Gateway surface:

- `gateway/stackchan_mcp/stdio_server.py` exposes `move_head(yaw, pitch)`.
- `move_head` has no `duration_ms` or speed parameter.
- It validates `yaw=-90..90`, `pitch=5..85`, then forwards unchanged arguments to firmware `self.robot.set_head_angles`.
- Firmware `self.robot.set_head_angles` also has no duration parameter, so user motion always falls back to the firmware default duration.

## Reference findings

- M5Stack official `m5stack/StackChan` and `m5stack/StackChan-BSP` use a custom spring-based `Servo`/`Motion` abstraction at 50 Hz / 20 ms. They still write each SCS servo separately with `WritePos(id, mapped_angle, 20, 0)`. They do not use ServoEasing for the SCS0009 path and do not use sync write in the inspected path.
[...]

PR #285, opened 2026-06-11
fix(firmware): #252 clamp raw speed_dps below step-safe floor in WriteHeadAngles

### LINKED ISSUE #252: Add stall-safe floor for raw-integer speed_dps escape hatch in move_head
## Context

PR #251 (Issue #129) introduces an LLM-friendly speed preset API for `move_head`. The gateway resolves `"low"` / `"mid"` / `"high"` presets to `speed_dps` integers (30 / 120 / 240) and forwards them to the firmware. Callers can also pass a raw integer (1..10000) as an escape hatch for fine-tuning.

The firmware-side `WriteHeadAngles(yaw, pitch, speed_dps)` overload clamps high values to `MAX_SPEED_DPS = 240` (SCS0009 datasheet endurance ceiling) and warns (without clamping) on values below `MIN_SMOOTH_SPEED_DPS = 72` (smoothness floor). This deliberately permits the `"low"` preset (30 dps) to deliver textured-but-moving slow motion.

## Problem

The raw-integer escape hatch has no lower-bound guard for a separate "stall floor":

- At `MOTION_TICK_MS = 20 ms`, one tick must advance at least one servo step to avoid sub-step stalls.
- SCS0009 step resolution = `0.293°` (300° / 1024).
- Minimum step-safe speed = `0.293° / 20 ms ≈ 14.65°/s`.

If a caller (typically via the raw-int escape hatch) passes `speed_dps ≤ 14`, the firmware computes a `duration_ms` that drives interpolation at less than one step per tick. The result is a sequence of `WritePos` calls with the same target held across multiple ticks before the servo advances by one step — the "sub-step stall" pattern characterized in Issue #129's design comment ("Min smooth speed ≈ 14.65°/s → below the floor → matches the on-device 'stepped' observation that originated #134").

This stall zone is **below** the smoothness-floor warning zone (72 dps). The smoothness floor is "looks textured but moves continuously"; this is "movement looks stuck / hangs briefly".

The preset paths (`"low"` = 30 dps, `"mid"` = 120 dps, `"high"` = 240 dps) all sit safely above this stall floor; the gap is only on the raw-int escape path.

## Scope

- Out of scope for PR #251 (kept clean for the preset-experience landing).
- Discovered during PR #251 review discussion.

## Proposed approach

1. Introduce `MIN_STEP_SAFE_SPEED_DPS = 15` `constexpr` in `firmware/main/boards/stackchan/stackchan.cc` next to the existing `MIN_SMOOTH_SPEED_DPS = 72` / `MAX_SPEED_DPS = 240` constants. Document the origin (`0.293° / MOTION_TICK_MS × 1000`, rounded up slightly for headroom).
2. In `WriteHeadAngles(yaw, pitch, speed_dps)`, clamp `safe_speed` up to `MIN_STEP_SAFE_SPEED_DPS` with a one-shot `ESP_LOGW` when `0 < speed_dps < MIN_STEP_SAFE_SPEED_DPS`. The smoothness-floor warn-only branch (`< MIN_SMOOTH_SPEED_DPS`) stays as-is.
3. Optional: align the gateway raw-int schema lower bound to `15` so the type-validation layer rejects sub-stall values before they reach the firmware.

## Acceptance

- Raw int `speed_dps` in `1..14` no longer causes sub-step stalls on real device (either clamps to 15 with a one-shot `ESP_LOGW`, or is rejected at the gateway schema layer).
- Preset paths (`"low"` / `"mid"` / `"high"`) behave identically to before; `"low"` = 30 stays above the new floor, no regression.
- Unit test covers the new clamp path on the gateway and/or firmware side.

## Priority

Low. The raw-int escape hatch is expected to be rarely used in practice (LLM clients default to presets); the immediate priority is the preset-experience landing in PR #251. This Issue captures the stall-floor gap so it is not lost, but does not block PR #251 merge.

## Refs

- Issue #129 (LLM-friendly speed preset API, parent design)
- PR #251 (initial preset implementation, parent PR)
- #134 (original boot-init "stepped" observation that surfaced the smoothness floor)

kisaragi-mochi:
## Summary

Closes the sub-step stall gap on the raw-integer `speed_dps` escape hatch of `move_head` / `WriteHeadAngles`.

The preset speeds ("low" = 30, "mid" = 120, "high" = 240 °/s) all sit above the physical step floor, but a raw integer in `(0, 15)` °/s drives the 20 ms motion tick below one SCS0009 step (300° / 1024 = 0.293°) per tick: interpolation then re-sends the same target across multiple ticks before the servo advances one step — the "movement looks stuck" pattern characterized in Issue #129's design discussion and first observed under #134.

## What changed

- `firmware/main/boards/stackchan/stackchan.cc`
  - New `MIN_STEP_SAFE_SPEED_DPS = 15` constant next to the existing speed constants, with the derivation documented (0.293° / 20 ms ≈ 14.65 °/s, rounded up for headroom) and a cross-reference to `BOOT_INIT_TARGET_DEG_PER_SEC` (#121/#141) — same physical origin, intentionally kept separate because the boot path carries its own duration-floor semantics.
  - `WriteHeadAngles(yaw, pitch, speed_dps)`: positive values below the floor are clamped up with one `ESP_LOGW` reporting requested and clamped values. The `<= 0` → default path and the warn-only smoothness band `[15, 72)` are unchanged.
  - The `self.robot.set_head_angles` MCP tool description notes the new lower-bound clamp.
- `CHANGELOG.md`: `[Unreleased]` → `### Firmware` entry.

## Verification

- Pre-PR review: clean (no findings).
- ESP-IDF Docker build (`release.py stackchan`): artifacts produced (`xiaozhi.bin` / `merged-binary.bin`).
- Real-device motion check at boundary speeds (e.g. raw 14 vs 15 vs 30) is deferred to the next on-device session; the change is conservative (raises only the stall band, never lowers speed).

## Refs

Closes #252.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ documented-only.
 
 - Added standalone CMake host-test infrastructure for firmware pure C++ helpers,
   covering mDNS gateway candidate extraction edge cases. (#279)
+- Clamp positive `speed_dps` values below 15 dps to the step-safe floor in
+  `WriteHeadAngles`, avoiding sub-step servo interpolation while preserving the
+  existing slow-motion preset behavior. (#252)
 
 ## [0.10.0] - 2026-06-09
 
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -796,6 +796,14 @@ private:
     static constexpr uint32_t MOTION_TICK_MS = 20;
     static constexpr uint32_t MOTION_DEFAULT_DURATION_MS = 600;
     // Speed-based motion API (Issue #129).
+    // MIN_STEP_SAFE_SPEED_DPS prevents the raw-integer speed_dps escape hatch
+    // from advancing less than one SCS0009 step per ServoTask tick. The
+    // physical step is 300 deg / 1024 = 0.293 deg; at MOTION_TICK_MS=20 ms
+    // this is 14.65 deg/s, rounded up to 15 deg/s for headroom. This shares
+    // the same physical origin as BOOT_INIT_TARGET_DEG_PER_SEC (#121/#141),
+    // but stays separate because the boot path carries its own duration-floor
+    // semantics. See the Issue #129 -> #134 stepped-motion observation lineage.
+    static constexpr int MIN_STEP_SAFE_SPEED_DPS = 15;
     // MIN_SMOOTH_SPEED_DPS is the on-device measured smoothness floor
     // (5 step/tick "transition out", measured 2026-05-15). Speeds below
     // this look textured on SCS0009 at MOTION_TICK_MS=20 ms; the firmware
@@ -3668,6 +3676,10 @@ private:
         int safe_speed = speed_dps;
         if (safe_speed <= 0) {
             safe_speed = DEFAULT_SPEED_DPS;
+        } else if (safe_speed < MIN_STEP_SAFE_SPEED_DPS) {
+            ESP_LOGW(TAG, "WriteHeadAngles: speed_dps=%d below MIN_STEP_SAFE_SPEED_DPS=%d, clamping",
+                     speed_dps, MIN_STEP_SAFE_SPEED_DPS);
+            safe_speed = MIN_STEP_SAFE_SPEED_DPS;
         } else if (safe_speed < MIN_SMOOTH_SPEED_DPS) {
             // Below the on-device measured smoothness floor -- motion will look
             // textured on SCS0009 at MOTION_TICK_MS=20 ms. This is intentionally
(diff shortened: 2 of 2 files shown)


kisaragi-mochi:
## Real-device verification (owner-realdevice-verify)

Verified on M5Stack CoreS3 + SCS0009 servos, firmware built from this branch (rebased on latest main, d2e88ff).

Test sequence — `move_head` yaw sweeps (30–60°) at raw `speed_dps` values targeting the clamp boundary:

| speed_dps | expected path | result |
|---|---|---|
| 1 | clamped to the 15 dps floor | completed, no stall, landed within ±1° |
| 10 | clamped to the 15 dps floor | completed, no stall |
| 14 | clamped to the 15 dps floor (upper boundary of clamp range) | completed, no stall |
| 15 | passes through unclamped (floor boundary) | visually identical to the clamped cases |
| `low` preset (30) | preset path, unaffected by the clamp | existing slow-preset behavior preserved |

Observations:

- All five motions completed without sub-step stalling; angle readbacks landed within ±1° of targets, with no failure-decoded values.
- Speeds 1–14 are visually indistinguishable from 15 dps, confirming the floor clamp engages as designed.
- At the 15 dps floor the motion shows a visible stepped cadence — consistent with the 0.293°/20 ms interpolation tick that defines the step-safe floor. The 30 dps preset appears visually smooth, so the smoothness threshold sits somewhere between 15 and 30 dps. Noted as a possible future enhancement (finer interpolation tick for smoother very-slow motion); not a blocker for this fix, which targets the stall.

Verdict: the clamp resolves the sub-step stall as intended. Verification hold can be lifted.

PR #317, opened 2026-06-27
Add edge-tts engine for natural English/multilingual TTS

scuddyracing:
VOICEVOX is Japanese-only by default, so non-Japanese text comes out heavily accented when spoken through it. This adds `edge-tts` (Microsoft's free online TTS service, via the `edge-tts` CLI) as an alternative engine for the `say` MCP tool.

## What this adds

- `EdgeTTSEngine` in `gateway/stackchan_mcp/tts/edge_tts.py`, following the same `TTSEngine` interface as the existing VOICEVOX/Irodori engines
- Registered as `"edge-tts"` in `tts/__init__.py`, behind the same `_try_register` ImportError guard used for the other engines

## How it works

- Shells out to the `edge-tts` CLI (no local model, no Docker container, no server process)
- `edge-tts` always emits MP3 regardless of the `--write-media` filename/extension passed, so this engine pipes that MP3 through `ffmpeg` to get raw 16 kHz mono PCM, matching the contract other engines follow
- Defaults to `en-GB-SoniaNeural`; overridable globally via `STACKCHAN_EDGE_TTS_DEFAULT_VOICE` or per-call via the `say()` tool's `voice`/`speaker_id` args (e.g. `en-US-AriaNeural`)

## Setup

```
pip install edge-tts
```
plus `ffmpeg` on PATH. No extras group changes needed since both are external CLI tools, not Python deps with native extensions.

## Testing

Manually tested end-to-end against a real M5Stack CoreS3 running this firmware: `say(text=..., voice="edge-tts")` synthesizes via the CLI, converts via ffmpeg, encodes to Opus, and plays correctly over the device speaker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,16 @@ documented-only.
   "downsample_hz must be a number in (0, 20]". The schema default
   (20) is unchanged, so calls that omit the argument behave the
   same. (#315)
+- Added `edge-tts` TTS engine — a subprocess-based engine using
+  Microsoft's Edge TTS CLI and ffmpeg, registered as `edge-tts`.
+  Provides natural English and multilingual voices (`voicevox` is
+  Japanese-only by default). Setup: install `edge-tts` CLI on PATH
+  (e.g. `pip install edge-tts`) plus `ffmpeg`. Default voice is
+  `en-GB-SoniaNeural`, overridable per-call via the new
+  `say(speaker_name=...)` argument (added to the say() schema as a
+  string field, distinct from `speaker_id` integer and the `voice`
+  engine selector) or globally via the `STACKCHAN_EDGE_TTS_DEFAULT_VOICE`
+  env var. (#317)
 
 ## [0.12.0] - 2026-06-20
 
--- a/gateway/stackchan_mcp/stdio_server.py
+++ b/gateway/stackchan_mcp/stdio_server.py
@@ -1606,8 +1606,20 @@ def create_server(notify_config: NotifyConfig | None = None) -> StackChanServer:
                         "speaker_id": {
                             "type": "integer",
                             "description": (
-                                "Engine-specific speaker identifier "
+                                "Engine-specific numeric speaker identifier "
                                 "(e.g. a VOICEVOX speaker ID)."
+                            ),
+                        },
+                        "speaker_name": {
+                            "type": "string",
+                            "description": (
+                                "Engine-specific string speaker/voice "
+                                "identifier (e.g. an Edge TTS voice name "
+                                "such as 'en-US-AriaNeural'). Distinct from "
+                                "'voice', which selects the engine itself "
+                                "(e.g. 'edge-tts'); use speaker_name for "
+                                "engines whose speaker selector is a string "
+                                "name rather than the numeric speaker_id."
                             ),
                         },
                         "reference_audio": {
--- a/gateway/stackchan_mcp/tts/__init__.py
+++ b/gateway/stackchan_mcp/tts/__init__.py
@@ -54,8 +54,15 @@ def _register_irodori() -> None:
     get_registry().register(IrodoriEngine())
 
 
+def _register_edge_tts() -> None:
+    from .edge_tts import EdgeTTSEngine
+
+    get_registry().register(EdgeTTSEngine())
+
+
 _try_register(_register_voicevox, "voicevox")
 _try_register(_register_irodori, "irodori")
+_try_register(_register_edge_tts, "edge-tts")
 
 
 __all__ = [
--- a/gateway/stackchan_mcp/tts/edge_tts.py
+++ b/gateway/stackchan_mcp/tts/edge_tts.py
@@ -13,6 +13,17 @@ get raw 16 kHz mono PCM.
 This is intentionally a subprocess-based engine (not an HTTP client like
 VoicevoxEngine) since edge-tts ships as a CLI tool, not a long-running
 server.
+
+Configuration (environment variables):
+
+    STACKCHAN_EDGE_TTS_DEFAULT_VOICE
+        Voice name used when the say() tool does not specify one via
+        speaker_name. Default "en-GB-SoniaNeural". Read lazily on each
+        access (not cached at import/construction time) so a .env file
+        loaded after engine registration — e.g. under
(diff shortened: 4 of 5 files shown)


kisaragi-mochi:
## Thank you for the contribution

Thanks for adding `EdgeTTSEngine` — multilingual / natural-English coverage is a real gap for users running stackchan-mcp outside Japanese-only use cases, and a CLI-based engine (no extra server, no API key) fits the existing "lightweight engines" pattern in `gateway/stackchan_mcp/tts/` well.

Real-device end-to-end testing on the M5Stack CoreS3 (per your PR description) is appreciated and lands the engine cleanly on the audio pipeline.

## Internal review findings

Three findings to address before merge, plus one carved out to a separate Issue.

### 1. `say()` schema vs Edge voice naming — `gateway/stackchan_mcp/tts/edge_tts.py:90`

The `say` MCP tool's schema declares `speaker_id` as an integer (matching VOICEVOX's integer speaker IDs). Under `serve --transport streamable-http`, the schema is enforced by jsonschema validation, so a string Edge voice name like `speaker_id: "en-US-AriaNeural"` would be rejected before reaching the engine. Conversely, if a number is sent (per schema), `EdgeTTSEngine.synthesize()` would pass that number as `--voice <integer>` to the `edge-tts` CLI and fail.

Suggested fix: add a string voice option (e.g. `voice: str`) to the `say` tool schema for Edge TTS use, or otherwise separate the types so the documented usage (`say(text=..., voice="en-US-AriaNeural")` from the PR body) reaches the engine through the HTTP transport.

### 2. `ffmpeg` subprocess inheriting MCP stdin — `gateway/stackchan_mcp/tts/edge_tts.py:113-119`

Under the default stdio MCP transport, the `ffmpeg` subprocess inherits the parent process's stdin — which is the MCP JSON-RPC pipe. `ffmpeg` reads `q` (and other interactive commands) from stdin, so if the MCP client sends a ping / cancellation / next request while `ffmpeg` is running, those bytes can be consumed by `ffmpeg`, potentially breaking the MCP session.

Suggested fix: pass `-nostdin` to the `ffmpeg` invocation and set `stdin=asyncio.subprocess.DEVNULL` in `create_subprocess_exec`, decoupling the child from the MCP stdin pipe.

### 3. `.env` default voice resolved too early — `gateway/stackchan_mcp/tts/edge_tts.py:63-64`

`STACKCHAN_EDGE_TTS_DEFAULT_VOICE` is read in `EdgeTTSEngine.__init__`, which runs at engine-registration (import) time. Under `serve --transport streamable-http`, the import chain `http_server -> stdio_server -> tts` resolves before `_configure_gateway_startup()` loads `.env`, so `.env`-based overrides for the default voice never take effect (the engine always uses `en-GB-SoniaNeural`).

Suggested fix: follow the pattern in `tts/irodori.py` — store only the explicit constructor override in `__init__`, and resolve the environment variable lazily inside `synthesize()` / `default_voice` so that a `.env` loaded later still applies.

### 4. Subprocess kill on timeout — carved out to #318

The `asyncio.wait_for(proc.communicate(), ...)` calls cancel the surrounding coroutine but don't terminate the spawned `edge-tts` / `ffmpeg` process on timeout, so processes can leak past tempdir cleanup if the upstream service hangs. This is low-priority (reproduction is rare) and orthogonal to the primary engine path, so it's tracked separately in #318 for follow-up — please don't feel obligated to address it within this PR.

## Wrap-up

Findings 1-3 are the ones we'd ask you to address before merge. Happy to keep this PR open while you iterate, and please feel free to ask if any of the suggestions are unclear or you'd prefer a different shape for the fix.

scuddyracing:
Thanks for the thorough review — all three points addressed in 456d61d.

**1. `speaker_id` vs Edge voice naming (`edge_tts.py:90`)**

Added a dedicated `speaker_name` string field to the `say` tool schema, kept separate from `speaker_id` (still `integer`, VOICEVOX-style) rather than loosening `speaker_id`'s type — `voice` was already taken as the engine selector, so reusing it for the speaker/voice-name string would have created a second collision. `speaker_name` is threaded through `orchestrator.py`'s `arguments.get(...)` extraction and the `engine.synthesize()` call, and `EdgeTTSEngine.synthesize` now reads `opts.get("speaker_name")` instead of misusing `speaker_id`/`voice`. Updated usage: `say(text=..., voice="edge-tts", speaker_name="en-US-AriaNeural")`.

**2. `ffmpeg` subprocess inheriting MCP stdin (`edge_tts.py:113-119`)**

Added `stdin=asyncio.subprocess.DEVNULL` to both the `edge-tts` and `ffmpeg` `create_subprocess_exec` calls, plus `-nostdin` on the `ffmpeg` invocation as suggested. Neither child can read from the MCP JSON-RPC pipe now.

**3. `.env` default voice resolved too early (`edge_tts.py:63-64`)**

Followed `irodori.py`'s pattern exactly: `__init__` now only stores `self._default_voice_override` (the explicit constructor arg), and `STACKCHAN_EDGE_TTS_DEFAULT_VOICE` is read lazily inside a `default_voice` property on every access, falling through to `DEFAULT_EDGE_TTS_VOICE` if unset. No more import-time env resolution.

**4. Subprocess kill on timeout (#318)**

Noted, and appreciate the heads-up that this is tracked separately — agreed it's orthogonal to this PR.

Re-tested end-to-end on the same M5Stack CoreS3 after all three fixes: `say(speaker_name="en-US-AriaNeural", ...)` now correctly selects a distinct voice (confirmed audibly, Aria vs the default Sonia) through the full schema → orchestrator → engine path, and the gateway starts and reconnects cleanly with no stdin-related session issues observed.

Let me know if you'd like anything else adjusted before merge.

kisaragi-mochi:
Round 2 review complete — thanks for the fast turnaround.

Findings 1/2/3 are all resolved. `pytest` (605 passed / 5 skipped),
`ruff check` on the changed files, and `codex review --base main` come
back clean modulo the subprocess-cleanup item already tracked in #318
(the engine's primary path is unaffected; #318 stays low-priority as a
carve-out, with action deferred to reactive triggers — user reports or
follow-up PRs — per maintainer judgement).

CHANGELOG entry added in `f7582a2` to satisfy the `CHANGELOG check`.
All CI checks green. Ready for merge.

PR #319, opened 2026-06-29
feat(firmware/i2c): add optional scl_speed_hz to Port A I2C tools

maha0525:
## What

Adds an optional `scl_speed_hz` property (default `400000`, range `100000`–`1000000`) to the three generic Port A I2C tools — `self.i2c.read`, `self.i2c.write`, `self.i2c.write_read`. When the property is omitted, behaviour is unchanged (400 kHz).

## Why

The tools currently register the Port A device handle at a hardcoded 400 kHz. Some M5Stack Units cannot sustain that clock. The **RCWL-9620 ultrasonic ranger** (Ultrasonic Distance Unit I2C, addr `0x57`) is a concrete case: it ACKs an `i2c_master_probe` (so `self.i2c.scan` lists it) but the actual transfer fails with `ESP_ERR_INVALID_STATE` at 400 kHz — the same failure mode the on-board PY32 IO expander hits at the wrong clock (see the existing comment near `Py32IoExpander`).

Measured on hardware (RCWL-9620 on Port A via PaHUB):

- `self.i2c.write` (trigger, master-driven) ACKs at 200 kHz
- `self.i2c.read` (measurement, slave-driven) is erratic at 200 kHz and only stable at **100 kHz** (5/5 identical readings)

This matches an independent community report of the same Unit returning garbage / 4500 mm at higher speed, fixed by dropping to 100 kHz: https://community.m5stack.com/topic/4255/ultrasonic-i2c-problems/2

Without a per-transaction clock override there is no way to drive such Units without recompiling the firmware.

## Change

`scl_speed_hz` is read from the property (default 400 kHz) and used for the per-call `i2c_master_bus_add_device` config in all three tools. No behavioural change when the property is omitted.

Built into an ESP-IDF 5.5 / esp32s3 firmware and verified on hardware: the RCWL-9620 reads stably at `scl_speed_hz: 100000`, while the default 400 kHz returns `ESP_ERR_INVALID_STATE`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,15 @@ documented-only.
 
 ### Firmware
 
+- Added an optional `scl_speed_hz` property (default `400000`, range
+  `100000`–`1000000`) to the Port A I2C tools `self.i2c.read`,
+  `self.i2c.write`, and `self.i2c.write_read`, applied to the per-call
+  `i2c_master_bus_add_device` config. This lets slow Units that cannot
+  sustain the hardcoded 400 kHz clock be driven without recompiling the
+  firmware (e.g. the RCWL-9620 ultrasonic ranger, which ACKs a probe but
+  fails the transfer with `ESP_ERR_INVALID_STATE` at 400 kHz and only
+  reads stably at 100 kHz). Behaviour is unchanged when the property is
+  omitted. (#319)
 - Added the NVS-backed `touch.enabled` field, handler guards, and
   `self.robot.set_touch_sensor_enabled` /
   `self.robot.get_touch_sensor_enabled` MCP tools so users can disable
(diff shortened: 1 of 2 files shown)


kisaragi-mochi:
Thanks for this — and for the careful hardware measurements backing it up. Maintainer review notes:

- The `Property("scl_speed_hz", kPropertyTypeInteger, 400000, 100000, 1000000)` five-arg overload resolves to (default, min, max), so omitting the property keeps the exact pre-PR behaviour (400 kHz hardcoded) — verified against `mcp_server.h` and the diff; no regression surface for existing setups.
- Tool descriptions mentioning the RCWL-9620 failure mode are a nice touch for LLM-facing discoverability.
- CI is green across all five checks.

One follow-up carved out on our side: the gateway wrappers (`i2c_read` / `i2c_write` / `i2c_write_read` in `stdio_server.py`) pass arguments through unchanged, so your property already works end-to-end — but their `inputSchema` doesn't declare `scl_speed_hz` yet, so schema-driven MCP clients can't discover it. Tracked in #321; we'll expose it on the gateway schemas in a follow-up PR. Nothing needed on your end.

LGTM — recommending merge as-is.

PR #320, opened 2026-07-01
fix(ownership): scope the gateway ownership lock per WS port

maha0525:
## What
Scope the gateway's ownership lock per WS port instead of a single machine-global `owner.lock` (path becomes `owner-<ws_port>.lock`, falling back to `owner.lock` when the WS port can't be resolved).

## Why
The lock guards against two processes fighting over the same device (`ownership.py`: `LOCK_PATH = LOCK_DIR / "owner.lock"`). The real invariant is **one gateway owns one device**; a machine-global lock over-encodes that as **one gateway per machine**. Running one gateway per device on separate WS ports on the same host makes the global lock wrongly reject the second gateway — it dies with `OwnershipError`, surfacing as `unhandled errors in a TaskGroup`. Scoping to the WS port restores the intended invariant: N gateways coexist as long as each owns a distinct port/device, while a duplicate start on the *same* port is still rejected.

## How
- `_ws_port_lock_path()` resolves to `owner-<ws_port>.lock` via existing `_resolve_ws_port()`, falling back to `owner.lock` when unresolved.
- `_acquire_startup_lock` passes `path=` to `acquire_lock` (stdio + streamable-http); release paths (atexit, stdio/http finally) use the same per-port path.
- `--check` reads the per-port lock, reflecting the port it would actually claim.
- `ownership.py` is unchanged — its functions already accept a `path` argument.

## Compatibility (single-device users unaffected)
A default user just sees `owner.lock` → `owner-8765.lock`; single-instance use and same-port double-start rejection are unchanged. The old `owner.lock` becomes a harmless orphan on upgrade.

## Verified on hardware
Two Stack-chan devices on one host (ports 18765 / 8767) start concurrently, each taking its own `owner-18765.lock` / `owner-8767.lock`; both independently controllable (head movement, speech, LEDs).

## Discussion
Happy to adjust the granularity — WS port is a proxy for "one gateway = one endpoint"; scoping by device identity (MAC / Device-Id) may fit better under a multi-device roadmap. Also open to adding cleanup for the orphaned legacy lock.

### DIFF
--- a/gateway/stackchan_mcp/cli.py
+++ b/gateway/stackchan_mcp/cli.py
@@ -774,7 +774,9 @@ def _ws_port_lock_path():
     from .ownership import LOCK_DIR
 
     ws_port, _ = _resolve_ws_port()
-    return LOCK_DIR / (f"owner-{ws_port}.lock" if ws_port else "owner.lock")
+    return LOCK_DIR / (
+        f"owner-{ws_port}.lock" if ws_port is not None else "owner.lock"
+    )
 
 
 def _acquire_startup_lock(
--- a/gateway/stackchan_mcp/cli.py
+++ b/gateway/stackchan_mcp/cli.py
@@ -438,9 +438,20 @@ def _run_ownership_check() -> int:
     """Print the current ownership lock status and exit cleanly."""
     from .ownership import is_pid_alive, read_lock
 
-    info = read_lock()
+    # Load ``.env`` first so that a WS_PORT / PORT defined only there resolves
+    # the same per-port lock the running gateway claims. Startup loads ``.env``
+    # before ``_acquire_startup_lock`` (via ``_run_stdio_gateway`` /
+    # ``_run_streamable_http_placeholder``), so ``--check`` must match it —
+    # otherwise a gateway owning ``owner-18765.lock`` would be inspected against
+    # the default ``owner-8765.lock`` and wrongly reported ready. Mirrors how
+    # ``--preflight`` loads ``.env`` inside ``_run_preflight``.
+    _load_dotenv()
+    # Inspect the per-WS_PORT lock this gateway would claim (not the legacy
+    # machine-global owner.lock), so the preflight reflects the actual port.
+    lock_path = _ws_port_lock_path()
+    info = read_lock(lock_path)
     if info is None:
-        print("no current owner")
+        print(f"no current owner (lock: {lock_path.name})")
         print("ownership preflight: ready")
         print("Result: ready. Exit 0.")
     elif is_pid_alive(info["pid"]):
@@ -750,6 +761,22 @@ def _configure_gateway_startup() -> None:
     log_user_defaults_startup()
 
 
+def _ws_port_lock_path():
+    """Per-WS_PORT ownership lock path.
+
+    複数機体対応: ownership lock を machine-global な ``owner.lock`` でなく
+    WS_PORT 単位に scope する。 lock は「同一 device を 2 つの gateway が
+    奪い合わない」 ための機構だが、 1 device = 1 gateway = 1 WS ポートで
+    複数機体を同時に動かす構成では global lock が 2 台目以降を誤って弾く。
+    ポート単位なら N gateway が共存でき、 同一ポート二重起動は従来通り弾ける。
+    WS_PORT 未解決時は従来の global パスにフォールバック。
+    """
+    from .ownership import LOCK_DIR
+
+    ws_port, _ = _resolve_ws_port()
+    return LOCK_DIR / (f"owner-{ws_port}.lock" if ws_port else "owner.lock")
+
+
 def _acquire_startup_lock(
     *,
     mode: "LockMode" = _STDIO_TRANSPORT,
@@ -764,13 +791,16 @@ def _acquire_startup_lock(
         release_lock_if_owner,
     )
 
+    lock_path = _ws_port_lock_path()
+
     owner_id = generate_owner_id()
     try:
         if mode == _STDIO_TRANSPORT and http_endpoint is None and started_by is None:
-            info = acquire_lock(owner_id)
+            info = acquire_lock(owner_id, path=lock_path)
         else:
             info = acquire_lock(
                 owner_id,
+                path=lock_path,
                 mode=mode,
                 http_endpoint=http_endpoint,
                 started_by=started_by,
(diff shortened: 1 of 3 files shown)


maha0525:
Pushed a revision addressing CI + the Codex review:

- **CHANGELOG check**: added a `### Gateway` entry under `[Unreleased]`.
- **Gateway test**: updated 4 `test_cli.py` mocks (`release_lock_if_owner` / `read_lock`) whose stubs assumed the pre-`path` call arity, so they now accept the per-port `path` argument the production code passes.
- **[P2] Load `.env` before choosing the check lock** (thanks): `_run_ownership_check` now calls `_load_dotenv()` before deriving the lock path, mirroring the startup order. Without it, a gateway configured via `.env` (e.g. `WS_PORT=18765`) would own `owner-18765.lock` while `--check` inspected the default `owner-8765.lock` and reported ready for the wrong port. Added a regression test (`test_main_check_flag_inspects_per_ws_port_lock`).

kisaragi-mochi (commented):
Thanks for the quick follow-up on the Codex finding — the `.env`-before-lock-path fix with its regression test looks good, and CI is green across the board. The per-port scoping itself is a clean cut, especially since `ownership.py` already took `path` everywhere.

Two small findings from the maintainer-side review — both are one-liners, and we're happy to fold them in on our end if you'd rather not spin another commit; just say the word.

**1. Dev-only test regression: the suite now breaks on a checkout with a real `gateway/.env` (CI can't see it)**

`test_main_check_flag_remains_side_effect_free_with_no_mdns` calls `main(["--check", "--no-mdns"])` without stubbing `_load_dotenv`. With this PR, `_run_ownership_check()` calls `_load_dotenv()` on that path, so the test loads the developer's real `gateway/.env` into `os.environ` — outside monkeypatch's tracking, so it is never restored. Subsequent tests then run in a polluted environment; with a `.env` that defines `STACKCHAN_TOKEN`, the six WS-server tests in `test_esp32_client.py` fail with `ESP32 auth rejected`:

```
uv run pytest tests/test_cli.py tests/test_esp32_client.py -q
# → 6 failed on a checkout with a real gateway/.env; green on CI (no .env there)
```

Minimal fix — stub the loader in that test the same way `_isolate_preflight_env` does:

```python
monkeypatch.setattr(cli, "_load_dotenv", lambda: None)
```

**2. `WS_PORT=0` resolves to the legacy lock** — left as an inline suggestion on the line.

On the Discussion points: WS-port scoping feels like the right granularity for now — in practice one gateway = one port = one device, and a device-identity (MAC / Device-Id) scheme only starts paying off once a broader multi-device story lands; happy to revisit it then. The orphaned legacy `owner.lock` is harmless, so leaving cleanup out of this PR is fine by us.

gateway/stackchan_mcp/cli.py
```diff
@@ -750,6 +761,22 @@ def _configure_gateway_startup() -> None:
...
+    WS_PORT 単位に scope する。 lock は「同一 device を 2 つの gateway が
+    奪い合わない」 ための機構だが、 1 device = 1 gateway = 1 WS ポートで
+    複数機体を同時に動かす構成では global lock が 2 台目以降を誤って弾く。
+    ポート単位なら N gateway が共存でき、 同一ポート二重起動は従来通り弾ける。
+    WS_PORT 未解決時は従来の global パスにフォールバック。
+    """
+    from .ownership import LOCK_DIR
+
+    ws_port, _ = _resolve_ws_port()
+    return LOCK_DIR / (f"owner-{ws_port}.lock" if ws_port else "owner.lock")
```
kisaragi-mochi: `_resolve_ws_port()` accepts `WS_PORT=0` as a valid ephemeral-port setting (and `--preflight` reports it as such), but the truthiness check here sends `0` to the legacy `owner.lock`. `is not None` keeps the fallback for genuinely unresolved values only. Admittedly a corner case — an OS-assigned port is unusual for this gateway since the device needs a stable URL — but the fix is one word:

```suggestion
    return LOCK_DIR / (f"owner-{ws_port}.lock" if ws_port is not None else "owner.lock")
```

maha0525:
Thanks for the thorough maintainer review — both were spot on, so I've pushed the two fixes in `94b8ef2` rather than have you fold them in.

**1. Test env leak.** Stubbed `_load_dotenv` in `test_main_check_flag_remains_side_effect_free_with_no_mdns`, mirroring `_isolate_preflight_env`. The root cause was a touch deeper than the loader itself: this PR switched `--check` from `_run_preflight` to `_run_ownership_check`, so the test's old `_run_preflight` stub no longer intercepted anything and `_run_ownership_check` → `_load_dotenv` ran for real. I reproduced the pollution locally with a `gateway/.env` carrying `STACKCHAN_TOKEN` — the WS-server suite went red with `ESP32 auth rejected`, green again after the stub.

**2. `WS_PORT=0` → legacy lock.** Applied the `is not None` fix so an ephemeral port scopes to `owner-0.lock`, consistent with how the preflight already treats `0` as valid. Added a regression test (`test_main_check_flag_scopes_lock_for_ephemeral_ws_port`) next to the existing per-port one.

Agreed on the discussion points too — WS-port granularity is the right granularity for now, and leaving the orphaned legacy `owner.lock` cleanup out of this PR is fine by me. Thanks again!

kisaragi-mochi:
Thanks for the quick turnaround — both fixes look great. The `_load_dotenv` stub with its root-cause note reads clearly (confirmed the full suite passes on a machine that has a real `gateway/.env`), and the ephemeral-port regression test is a nice addition.

Heads-up on the branch state: after your push, we cut the v0.13.0 / firmware-v1.13.0 release and merged two gateway PRs (#324, #325) on our end, which rewrote the CHANGELOG `[Unreleased]` section and put this branch into conflict. That was entirely on our side, so we've handled it here: rebased onto the latest main, moved your changelog entry into the current `[Unreleased] > Gateway` section, and force-pushed with your authorship intact. Nothing to do on your end — if you have local work on this branch, syncing with `git fetch` + `git reset --hard <your-remote>/pr-o-ownership-lock-per-port` will line it back up.

CI is now running against the rebased head; we'll proceed with the merge once it's green.

kisaragi-mochi (approved):
Round-2 fixes verified: the `_load_dotenv` stub and the `is not None` port scoping both look correct, the full suite is green locally (617 passed) including on an environment with a real `gateway/.env`, and CI is green across the board on the rebased head. LGTM.

PR #326, opened 2026-07-02
fix(capture): don't 413 uploads on aiohttp>=3.14; tolerate non-UTF-8 question

maha0525:
Two robustness fixes for the `/capture` photo-upload endpoint.

## Fix 1 — `/capture` returns 413 for normal uploads on aiohttp >= 3.14

`create_capture_app` sets `web.Application(client_max_size=0)` to lift aiohttp's default 1 MiB body cap so `/pcm` can stream arbitrarily long PCM.

aiohttp >= 3.14 added a size check to its multipart reader (`BodyPartReader.read()`: `if len(data) > client_max_size: raise HTTPRequestEntityTooLarge`) that — unlike `request.read()` / `request.post()`, which guard with `if 0 < max_size` — does **not** treat `0` as "unlimited". `request.multipart()` propagates the app's `client_max_size` (`0`) into that reader, so every `/capture` upload whose `question` form field is non-empty (read via `part.read()`) is rejected with **413 "Maximum request body size 0 exceeded"**. Empty-question uploads slip through (the file part is read via `read_chunk`), which masks the bug.

Downstream this surfaces as the ESP32 reporting **"Failed to upload photo"**: the device POSTs the JPEG, the gateway 413s it, `take_photo` returns an error.

**Fix:** use `client_max_size=sys.maxsize` instead of `0`. `request.read()` / `.post()` and multipart are all effectively unbounded, `/pcm` keeps streaming, and `/capture`'s real limit stays the explicit `CAPTURE_MAX_BYTES` per-route check in `handle_capture`.

Only manifests on aiohttp >= 3.14 (where the multipart size check was added); on 3.13.x the endpoint worked regardless of `client_max_size`.

## Fix 2 — non-UTF-8 `question` field 500s the upload

`handle_capture` decoded the `question` field with a strict `.decode("utf-8")`, so a client that sent it in another encoding turned a valid photo upload into a 500. The question is advisory metadata; decode with `errors="replace"` so the photo is still stored.

## Tests

- `test_capture_accepts_nonempty_question_field`: posts a non-empty `question` via `TestClient`; fails with 413 under the old `client_max_size=0` on aiohttp 3.14, passes now.
- `test_capture_tolerates_non_utf8_question`: posts raw non-UTF-8 `question` bytes; 500 without the decode change, 200 now.

Both verified against aiohttp 3.14.1 (which reproduces the 413) and on real hardware (two Stack-chan devices; the camera `see` path went from "Failed to upload photo" to returning the image).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,15 @@ documented-only.
   (the default lock is simply named `owner-8765.lock`) and a duplicate
   start on the same port is still rejected; `--check` reads the per-port
   lock. (#320)
+- Fixed `/capture` rejecting photo uploads with HTTP 413 on aiohttp >= 3.14.
+  The capture app raised the per-request body cap with `client_max_size=0`
+  (so `/pcm` can stream long PCM), but aiohttp's multipart reader treats `0`
+  as a zero-byte limit — unlike `request.read()`/`.post()` — so every upload
+  carrying a non-empty `question` field was rejected with 413. Use a large
+  finite `client_max_size` instead; `/capture`'s real limit stays the explicit
+  per-route byte cap.
+- Hardened `/capture` to tolerate a non-UTF-8 `question` field (decode with
+  `errors="replace"`) instead of failing the upload with HTTP 500.
 
 ## [0.13.0] - 2026-07-02
 
--- a/gateway/stackchan_mcp/capture_server.py
+++ b/gateway/stackchan_mcp/capture_server.py
@@ -143,7 +143,11 @@ async def handle_capture(request: web.Request) -> web.Response:
 
     async for part in reader:
         if part.name == "question":
-            question = (await part.read()).decode("utf-8")
+            # Decode defensively: a malformed (non-UTF-8) question field must
+            # not turn a photo upload into a 500. The question is advisory
+            # metadata, so replacing undecodable bytes is preferable to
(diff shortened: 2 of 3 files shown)


kisaragi-mochi:
Thanks for another well-diagnosed pair of fixes — the aiohttp >= 3.14 multipart cap analysis is spot on (nice catch that empty-question uploads masked the bug), and both regression tests reproduce the failure modes cleanly. The full suite is green locally (619 passed) and a Codex review came back clean as well.

Same branch-state note as #320: our v0.13.0 release and the gateway PRs merged since (#324, #325, #320) rewrote the CHANGELOG `[Unreleased]` section, which put this branch into conflict — that's on our side, not yours. We've rebased onto the latest main here, kept your two changelog bullets under the current `[Unreleased] > Gateway`, and force-pushed with your authorship intact. Nothing to do on your end.

CI is now running against the rebased head; we'll merge once it's green.

kisaragi-mochi (approved):
Both fixes verified: the `sys.maxsize` body-cap change keeps /pcm unbounded while /capture stays guarded by its explicit `CAPTURE_MAX_BYTES` route check, and the `errors="replace"` decode correctly saves the photo when the advisory question field is malformed. Full suite green locally (619 passed) on the rebased head, Codex review clean. LGTM.

Issue #328, opened 2026-07-02
Issue_listen_crash


jeaniezhao106:
## Describe the bug

Calling the `listen()` MCP tool causes the ESP32 to disconnect and reconnect, receiving zero audio frames.

## Environment

| Component | Version |
|-----------|---------|
| Gateway | v0.13.0 (pip install, streamable-http mode) |
| Firmware | v2.2.6 (stackchan board, firmware-v1.13.0 release) |
| Board | M5Stack StackChan (CoreS3 + Feetech servos + GC0308) |
| Host | macOS 14.6 (Apple Silicon M3), Python 3.12.13 |
| TTS engine | edge-tts (works fine) |
| STT engine | faster-whisper |

## Steps to reproduce

1. Start gateway: `stackchan-mcp serve --transport streamable-http`
2. ESP32 connects successfully, reports 32 tools, `get_status` works, `say` (TTS) works
3. Call `listen()` with any duration:

```json
{
  "method": "tools/call",
  "params": {
    "name": "listen",
    "arguments": {
      "duration_ms": 3000,
      "engine": "faster-whisper",
      "language": "zh"
    }
  }
}
```

## Expected behavior

ESP32 enters listening mode, streams Opus audio frames over WebSocket, gateway decodes and transcribes via faster-whisper, returns text.

## Actual behavior

ESP32 **disconnects from WebSocket** ~400ms after receiving `listen.start`, then reconnects ~40ms later with a fresh session. Gateway receives 0 Opus frames.

Result: `{"text": "", "frame_count": 0, "duration_ms": 0}`

## Gateway logs

```
ESP32 initialized: protocol=2024-11-05 server={'name': 'stackchan', 'version': '2.2.6'}
ESP32 ready: device=44:1b:f6:e1:e4:0c tools=32

# TTS works fine:
send_pcm_audio: source=engine:edge-tts frames=138 duration_ms=8280   ✅

# listen() triggers disconnect:
ESP32 disconnected: device=44:1b:f6:e1:e4:0c                          ❌
connection open
ESP32 connecting: device=44:1b:f6:e1:e4:0c
ESP32 initialized: protocol=2024-11-05 server={'name': 'stackchan', 'version': '2.2.6'}
ESP32 ready: device=44:1b:f6:e1:e4:0c tools=32
listen(): no Opus frames received during 3000 ms window
listen(): engine=faster-whisper frames=0 duration_ms=0 text=''
```

## Analysis

[...]

Issue #233, opened 2026-05-26
Touch-driven listen 30s auto-stop also stops non-touch-initiated listening sessions


kisaragi-mochi:
## Context

PR #208 introduces a touch-driven listen UX for the stack-chan board, including a 30-second auto-stop timer in `PollTouchpad` that calls `StopListening()` when a listening session is left idle (intended as a safety net for "user started a listen by touch, then walked away").

## Observation

The 30s timeout keys off `kDeviceStateListening` without checking how the device entered the Listening state. As a result, listening sessions initiated by paths other than a touch tap — e.g. wake-word activation or realtime listening — are also subject to the 30s auto-stop and would be terminated unexpectedly.

## Current impact

None in the current deployment. stack-chan is driven via touch / MCP, and wake-word configuration is managed upstream (not configurable on this side at present), so non-touch-initiated listening does not occur in practice.

## Future consideration

If wake-word configuration becomes available on this side, the 30s timeout should be scoped to touch-initiated sessions only (e.g. track whether the session was started from the touch branch) so hands-free listening is not cut off.

## Type

Tracking / observation only. No action required now. Surfaced by codex review (P2) on PR #208.

Issue #220, opened 2026-05-21
firmware: play_popup_on_listening_ flag leak on OpenAudioChannel failure (follow-up to #207)


kisaragi-mochi:
## Background

Follow-up to #207 (which introduced `play_popup_on_listening_` gating to bring the popup audio cue to non-wake-word activation paths). A second-pass codex review on the merged revision surfaced a state-leak corner case carved out from #207 itself, because the fix touches the broader state-machine contract for failed / aborted `OpenAudioChannel()` transitions.

## Problem

In `Application::HandleStartListeningEvent()` (`firmware/main/application.cc`, around line 788-803), `play_popup_on_listening_` is set when `state == kDeviceStateIdle || state == kDeviceStateSpeaking`.

In the `Idle && audio channel not yet open` branch, the flag is set BEFORE `ContinueOpenAudioChannel()` (which calls `OpenAudioChannel()`).

If `OpenAudioChannel()` fails (TLS handshake error, gateway unavailable, transient state change mid-handshake), or if the device state changes before the Listening transition completes, the flag remains armed.

The next unrelated Listening transition (e.g. a wake-word event after WiFi recovery) will then:

- Play the popup audio cue when the user did not initiate this activation
- Pass the stale gating into the `SendStartListening()` send-condition branch, contaminating the WebSocket protocol toward the gateway

## Reproduction

- WiFi unstable / association failure environments
- Tailscale Funnel TLS handshake intermittent failures (related: #102)
- Gateway restart during a touch-activated listen attempt

In stable LAN environments the failure path is rare and the leak window closes on the next state transition into Listening.

## Suggested fix

Either:

- Set `play_popup_on_listening_ = true` only after the Listening transition is confirmed (move into the success callback of `OpenAudioChannel()`), or
- Clear the flag on every failure / abort path of `ContinueOpenAudioChannel()`

The first option is preferred because it matches the wake-word path contract: the flag is set right before `ContinueWakeWordInvoke()`, and `HandleStateChangedEvent` clears it on transition out.

## Scope

Carved out from #207 to keep that PR scoped to the popup-cue API extension that was reviewed. The fix is a few lines but its design touches the broader state-machine contract for failure / abort transitions, so a focused follow-up PR is the right home.

This finding is independent of PR #208 (touch-driven listen UX), which uses the same `StartListening()` API path.

## Refs

- PR #207 (origin of the gating refactor): #207
- PR #208 (same `StartListening()` API consumer): #208
[...]

PR #330, opened 2026-07-05
fix(firmware/ws): arm audio session in place when the transport is already connected (#328)

kisaragi-mochi:
## Summary

Fixes the server-driven `listen()` failure reported in #328: calling the gateway's `listen()` MCP tool while the device sits idle caused the ESP32 to tear down its healthy WebSocket and reconnect, so the gateway's capture window (bound to the old connection) received zero audio frames.

`WebsocketProtocol::OpenAudioChannel()` now arms the logical audio session **in place** when the transport is already connected and the server hello has been acked, instead of unconditionally rebuilding the socket.

## Root cause

Confirmed analysis in #328 (see the issue comment for the full walkthrough):

- PR #136 made `CloseAudioChannel()` keep the WS alive; PR #192 made `audio_channel_open_` a logical-session flag, with the boot-time connect and every reconnect opening the socket via `OpenAudioChannelInternal(false, false)` — i.e. transport up, MCP working, audio session **not** armed.
- `HandleStartListeningEvent()` still followed the upstream assumption that "opening the audio channel" means "establishing the connection": from idle with `IsAudioChannelOpened() == false` it called `OpenAudioChannel()`, whose prologue does `websocket_.reset()` — destroying a healthy socket.
- `say()` (TTS) was unaffected because `SendAudio()` only checks transport connectivity, which is why TTS worked while `listen()` failed 100% of the time.

## Fix

`OpenAudioChannel()` first evaluates an `existing_transport_ready` predicate: `transport_connected_`, `websocket_` non-null, `websocket_->IsConnected()`, `!error_occurred_`, plus `!session_id_.empty()` as the hello-acked signal (PR #192 requires a non-empty `session_id` for a successful hello).

- Predicate true → refresh `last_incoming_time_` (the audio-session deadline clock starts at this open, matching the MQTT transport's `OpenAudioChannel()`), arm `audio_channel_open_` via `exchange(true)`, fire `on_audio_channel_opened_()` only on the false→true transition (keeps the power-save level transition symmetric with `OnAudioChannelClosed`), and return `true` without touching the socket.
- Predicate false (boot, dead socket, error, hello not acked) → unchanged full `OpenAudioChannelInternal(true, true)` path.

The predicate deliberately does **not** consult `Protocol::IsTimeout()`: that tracks the audio-session inbound-frame deadline (120 s since the last inbound message), while an idle persistent MCP socket can sit for minutes with no inbound frames and still be healthy transport — the same reasoning already documented on `IsTransportConnected()`. This was a round-1 internal review finding: gating on `IsTimeout()` would have skipped the fast path in the common "idle a few minutes, then listen/touch/wake" case, reintroducing the teardown this PR removes.

Touch and wake-word activations from the same idle state take the same shortcut, so those no longer rebuild the socket either.

## Behavior

| State when `OpenAudioChannel()` is called | Before | After |
|---|---|---|
| WS connected + hello acked, session not armed (post-boot idle) | socket teardown + rebuild; gateway sees disconnect | in-place arm; socket untouched |
| Session already armed | teardown + rebuild | no-op, returns `true` |
| No transport / dead socket / error / timeout / no hello | full rebuild path | unchanged |

## Verification

- Internal codex review: 2 rounds — round 1 surfaced the `IsTimeout()` gating issue (accepted, fixed in the second commit), round 2 clean.
- Firmware build green via `espressif/idf:v5.5.2` Docker (`release.py stackchan`): `xiaozhi.bin` 3.4 MB, `merged-binary.bin` 9.5 MB.
- `git diff --check` clean; change confined to `firmware/main/protocols/websocket_protocol.cc` + `CHANGELOG.md`.
- Real-device E2E (`listen()` from idle over a live gateway connection, confirming no disconnect and non-zero frame capture) is planned before merge; results will be posted on this PR.

## Out of scope

- #188 (late `OnDisconnected` callback ordering), #220 (`play_popup_on_listening_` leak on open failure), #233 (30 s auto-stop scoping) — adjacent known issues in the same area, tracked separately.
- Gateway-side changes (none needed).
- Threading `mode` through the server-driven listen message (follow-up noted in #91/#93).

Refs #328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ documented-only.
 
 ## [Unreleased]
 
+### Firmware
+
+- Fixed server-driven listen activation on an already connected WebSocket to arm
+  the logical audio session in place instead of rebuilding the transport, so the
+  gateway capture window remains bound to the live connection. (#328)
+
 ## [0.14.0] - 2026-07-03
 
 ### Gateway
--- a/firmware/main/protocols/websocket_protocol.cc
+++ b/firmware/main/protocols/websocket_protocol.cc
@@ -213,19 +213,23 @@ bool WebsocketProtocol::OpenAudioChannel() {
         websocket_ != nullptr &&
         websocket_->IsConnected() &&
         !error_occurred_ &&
-        !IsTimeout() &&
         !session_id_.empty();
     if (existing_transport_ready) {
         // Issue #328: PR #136 and PR #192 made the audio session a logical
         // state that can be closed while the WebSocket transport stays open.
         // Server-driven listen.start, touch, and wake-word activations from
         // that idle state should therefore arm the session on the current
-        // socket instead of rebuilding a healthy transport. The transport
-        // predicate mirrors IsAudioChannelOpened() without the logical flag,
-        // and a non-empty session_id_ is the ParseServerHello-acked signal.
+        // socket instead of rebuilding a healthy transport. Do not gate this
+        // transport predicate on Protocol::IsTimeout(): that tracks the
+        // audio-session inbound-frame deadline, while an idle persistent MCP
+        // socket can have no inbound frames for minutes and still be healthy.
+        // Refresh last_incoming_time_ when arming so IsAudioChannelOpened()
+        // starts its session deadline from this open, matching MQTT.
+        // A non-empty session_id_ is the ParseServerHello-acked signal.
         // session_id_ is read here on the main task following the existing
         // session-id gate pattern; broader cross-task hardening is out of
         // scope for this fix.
+        last_incoming_time_ = std::chrono::steady_clock::now();
         if (!audio_channel_open_.exchange(true) && on_audio_channel_opened_ != nullptr) {
             on_audio_channel_opened_();
         }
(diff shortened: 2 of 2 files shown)


kisaragi-mochi:
## Real-device E2E verification — PASS

Setup: this PR's firmware flashed to the CoreS3 (app partition only, NVS preserved), device auto-reconnected to a gateway via mDNS, then server-driven `listen()` was fired three times through the streamable-http MCP endpoint while the device sat idle (including a several-minutes-idle window, exercising the round-2 idle-timeout path).

| Run | Window | Frames received | `ESP32 disconnected` during listen |
|---|---|---|---|
| 1 | 10 s | 150 | none |
| 2 | 10 s | 164 | none |
| 3 | 15 s | 248 (non-empty transcription returned) | none |

Gateway log over the whole session shows exactly one connection event pair (`connecting` → `ready`) and **zero disconnects** — the same log pattern that reproduced 100% as disconnect + 0 frames on the pre-fix firmware (#328).

The full pipeline (`listen.start` → Opus frames → decode → faster-whisper → text) is functional. Transcription quality on quiet input is a separate known topic (#95 mic input gain, #299 STT accuracy refinement) and out of scope here.

CI is green (6/6, including the firmware build). Ready for merge review.

Issue #335, opened 2026-07-09
feat(gateway): follow_led_stream — drive device LEDs from a WebSocket frame stream


kisaragi-mochi:
## Summary

Add a gateway tool `follow_led_stream` — a sister tool of `follow_pose_stream` — that
connects to an external WebSocket URL as a client, receives a stream of LED frames, and
forwards them to the device LEDs (base 12-LED ring and/or Port B WS2812 strips).

## Motivation

During a recent live music session, the device's LEDs were driven by manually issuing
`set_leds` calls in time with the music. That works as a fallback but does not scale to
automated, beat-synchronized lighting. An external bridge process can convert musical
performance events into LED frames; the gateway only needs a generic ingestion endpoint
for those frames — the same separation `follow_pose_stream` uses (it consumes normalized
`{yaw, pitch}` frames, not raw IMU data). The event→frame conversion logic stays outside
this repo.

## Proposed design

Mirror the `follow_pose_stream` architecture 1:1 where it fits:

- Tool `follow_led_stream` with `action: start | stop | status`
- Gateway connects to the configured WS URL as a client, with exponential backoff
  reconnection and a module-level singleton guarded by a lock
- Upstream auth via URL query token (same pattern as pose stream; no header auth needed)

### Frame schema (public API)

```json
{
  "ts": 1751234567890,
  "kind": "event",
  "colors": [[255, 0, 0], [0, 0, 255]]
}
```

- `ts` — sender epoch ms (naming consistent with pose stream frames)
- `kind` — `"event"` or `"continuous"`:
  - `event` frames (e.g. a flash on a beat) bypass the downsample gate and are never
    silently dropped
  - `continuous` frames (e.g. a hue rotation effect) are downsampled to `max_fps`
  - Rationale: pose stream's single elapsed-time gate silently drops frames that arrive
    too fast, which is fine for head tracking but would eat rapid flash events
- `colors` — full RGB888 (0–255) array in LED index order; malformed payloads are
  rejected as a whole (same policy as the firmware `set_strip` handler)
- Optional `source` / `frame` fields for filtering, mirroring pose stream's
  `source_filter` / `frame_filter`

### Target selection

Config selects the output target:

- `base_ring` — the built-in 12-LED ring (`self.led.set_many` path; note RGB565
  quantization in firmware)
- `port_b` — Port B WS2812 strip (`self.port_b.ws2812.set_strip` path; RGB888 full
[...]

Issue #307, opened 2026-06-20
gateway/follow_pose_stream: hardening for ESP32 mid-stream reboot


kisaragi-mochi:
## Problem

`stackchan_follow_pose_stream` caches device state across the
session (initial seed position, WiFi PS apply result) so the gateway
can apply step-rate limiting against the live head position and
disable the DTIM-cycle WiFi power-save jitter.

These cached fields are invalidated on explicit `set_head_angles`
disconnect errors (PR #306 R10), but they are NOT invalidated when
the ESP32 cleanly disconnects + reconnects while no pose frame is
in flight or while incoming frames are dropped by the configured
filters. After such a reconnect a new device session has fresh
servo state (e.g. yaw=0, pitch=45 from boot), but the gateway
still holds the previous session's `_last_servo_*`. The first
command after reconnect therefore step-clamps against stale state
and can issue a large physical swing.

The WiFi PS apply state is also stale across the reconnect, so the
next stream window runs with the firmware default (modem sleep)
and the ~800 ms send jitter that the tool is designed to avoid
returns.

## Proposed fix direction

Inject the active ESP32 connection / session id into
`FollowPoseStream`. Snapshot it before the seed guard in
`_consume()` (and in the `_run()` seed path) and invalidate the
seed + WiFi PS apply state when the id changes between iterations,
so the next valid frame re-seeds from the live head pose and
re-applies WiFi PS="none" before issuing any servo command.

## Scope

- gateway-only (`follow_pose_stream.py` + `esp32_client.py`
  session-id exposure)
- no firmware change
- no public MCP tool surface change

## Related

- #304 — pose-stream subscriber design
- #306 — initial implementation (R6-R10 review iterations)

## Priority

priority/low — only triggers on mid-stream ESP32 reboot, which is
rare in normal use; the more common in-stream disconnect error
path is already handled in #306 R10.

Issue #224, opened 2026-05-22
feat(gateway): expose self.port_b.ws2812.* as MCP tools (follow-up to #223)


kisaragi-mochi:
## Background

PR #223 (merged 2026-05-22) added firmware-side
`self.port_b.ws2812.*` generic MCP tools — five tools that drive
any WS2812-compatible LED strip connected to the official kit's
Port B (CoreS3 HY2.0-4P digital OUTPUT, GPIO 9). These are
visible from the gateway's `get_status` call (`tools_count: 30`,
including the five `self.port_b.ws2812.*` entries), and the
firmware accepts the calls correctly.

However, the **gateway-side MCP wrapping** that would let an
LLM client invoke these tools is **not yet implemented**. The
`tool_map` in `gateway/stackchan_mcp/stdio_server.py` (around
L915) only contains the 27 hardcoded mappings present before
PR #223. A downstream MCP client call to any
`self.port_b.ws2812.*` tool currently returns
`{"error": "Unknown tool: ..."}`.

This is the same pattern as PR #196 (Port A I2C generic), where
both firmware and gateway side wrapping landed together. PR #223
was firmware-only; this Issue tracks the matching gateway-side
follow-up.

## Scope

Add five new gateway-side MCP tools, each forwarding the call to
the corresponding firmware tool through `gw.esp32.call_tool(...)`:

| Gateway MCP tool name | Firmware tool name | Arguments | Notes |
|---|---|---|---|
| `port_b_ws2812_init` | `self.port_b.ws2812.init` | `led_count: int (1..256)` | Allocates the `led_strip` RMT backend. Repeated calls with the same `led_count` are no-ops; a different `led_count` tears down and rebuilds the strip handle. |
| `port_b_ws2812_set_pixel` | `self.port_b.ws2812.set_pixel` | `index: int, r: int (0..255), g: int (0..255), b: int (0..255), refresh: bool (default false)` | Sets one LED in the buffer; `refresh=true` auto-latches. |
| `port_b_ws2812_set_strip` | `self.port_b.ws2812.set_strip` | `colors: list[list[int]]` | Sets multiple LEDs and auto-refreshes. Validate-then-write semantics. |
| `port_b_ws2812_refresh` | `self.port_b.ws2812.refresh` | (none) | Latches the buffered colors out. |
| `port_b_ws2812_clear` | `self.port_b.ws2812.clear` | (none) | Clears the strip. |

[...]

Issue #73, opened 2026-05-09
feat(mcp): parallel dispatch for hardware-independent tools (servo + LED concurrency)


kisaragi-mochi:
## Problem

When agents call hardware-affecting tools concurrently (e.g. `move_head` together with `set_leds` for an animated demo), LED frames visibly stall during servo motion — even though servo (UART) and LED (I2C via the PY32 IO expander) drive completely independent peripherals.

Observed during a real-device demo run on the PR #68 LED tools. A sequence that interleaved `move_head` calls between rotating-rainbow `set_leds` frames showed visibly slower / stuttery LED motion compared to the same LED frames issued without `move_head` in between.

## Hypothesis

The on-device MCP server dispatcher appears to serialize incoming `tools/call` messages: the next message is not handled until the current handler returns. `move_head` returns quickly (it just sends a UART command and the servo runs asynchronously in hardware), but any per-message overhead — UART TX, optional register readback, context switching — adds enough latency to perceptibly delay subsequent LED I2C writes that would otherwise pipeline cleanly.

Even if each handler is individually fast, **serialization across hardware-independent tools wastes the natural parallelism of separate ESP32 peripherals**.

### A. Parallel dispatch with hardware-category tags
Tag each tool with a "hardware lane" (e.g. `servo`, `led`, `display`, `audio`, `camera`). The dispatcher serializes within a lane (multiple `move_head` calls genuinely need ordering), but allows different lanes to run in parallel via separate FreeRTOS tasks.

- ✅ Pro: clean general solution; future tools just declare their lane.
- ⚠ Con: dispatcher complexity; lane interactions need care (e.g. display + LED can share I2C / SPI buses depending on board).

### B. Combined "compound" tools
Add explicit MCP tools that bundle hardware-independent operations:

```
move_with_leds(yaw, pitch, leds)
move_with_avatar(yaw, pitch, face)
```

These run servo + LED / display in parallel inside a single handler.

- ✅ Pro: simplest; very predictable.
- ⚠ Con: surface area grows quickly as more combinations are needed.

### C. Move long-running handler logic to a background task
Keep the dispatcher serial, but make individual handlers (especially `move_head`) hand off to a FreeRTOS task and return immediately. The next message arrives sooner.

- ✅ Pro: minimal dispatcher changes.
- ⚠ Con: still doesn't truly parallelize; just lowers per-handler dispatch latency.

## Recommendation

[...]

PR #337, opened 2026-07-09
feat(gateway): follow_led_stream — WS LED frame streaming (base ring / Port B) (#335)

kisaragi-mochi:
## Summary

Implements `follow_led_stream` (#335) — a sister tool of `follow_pose_stream` that
connects to an external WebSocket URL as a client, receives LED frames, and forwards
them to the device LEDs.

- **Frame schema (public API)**: `{"ts": <epoch ms>, "kind": "event"|"continuous", "colors": [[r,g,b], ...]}`
  with optional `source` / `frame` filter fields. Malformed frames are rejected as a
  whole (same policy as the firmware `set_strip` handler).
- **Event vs continuous**: `continuous` frames are downsampled to `max_fps`
  (default/cap 30); `event` frames bypass the gate so beat flashes are never silently
  dropped.
- **Target selection**: `base_ring` (built-in 12-LED ring via `self.led.set_many`) or
  `port_b` (WS2812 strip via `self.port_b.ws2812.set_strip`, auto-`init` on start with
  the configured `led_count`).
- **Fallback semantics**: no exclusivity mechanism — while the stream is stopped or the
  WS client is disconnected, nothing is sent to the device, so manual `set_leds` /
  `port_b_ws2812_*` calls keep working with no switching step.
- **Device-reset recovery** (review finding): an `available=false` reply on `port_b`
  dispatch is treated as a device reset — device state is invalidated (re-arming init +
  WiFi power-save re-apply) and the same frame is re-dispatched exactly once.

### Related fixes in scope (per #335)

- **Shared refcounted WiFi power-save helper** (`wifi_power_save.py`):
  `follow_pose_stream` and `follow_led_stream` now acquire/release a shared lease —
  when both streams run concurrently, one stopping no longer restores power save under
  the other. `follow_pose_stream` refactored to use the helper (behavior unchanged when
  running alone).
- **Dedicated tool lane for `self.port_b.`** in `_TOOL_LANES`, so strip writes
  serialize against each other instead of falling through to the default lane.
- LED follower registered in the gateway shutdown path (mirrors the pose follower).

## Out of scope

- The external bridge that converts music/performance events into LED frames (lives
  outside this repo; depends only on the frame schema above).
- Firmware changes — none; the wire-level LED tools already landed (#223, #224, #291).
- On-device latency measurement (the Issue's acceptance latency gate) — planned as the
  real-device verification step before this feature is relied on in performance use.

## Validation

- `cd gateway && uv run pytest -q` → 661 passed, 1 skipped (includes 30+ new tests:
  config/frame validation, event-bypass vs continuous gating, per-target dispatch
  payloads, port_b init/reset recovery, singleton semantics, WiFi power-save refcount
  across two concurrent streams)
- `uv run ruff check .` → clean
- Codex review: 3 rounds — 2 P2 findings (gateway-shutdown cleanup, port_b device-reset
  recovery) fixed in rounds 1-2; round 3 clean
- CHANGELOG `[Unreleased] Gateway` entry added
- README.md / README.ja.md both updated (dual-language sync)

## Refs

- Refs #335 (close after real-device verification)
- Design references: #69 (complementary firmware-side animation), #73 (tool lanes), #307 (pose stream hardening)


kisaragi-mochi:
## Real-device verification report

Setup: gateway running this branch on the host, ESP32 (CoreS3) over home WLAN,
target `base_ring`, dummy frame server feeding the documented schema at 30 fps
continuous (hue rotation) plus a 3-frame `event` burst (white flash) every 3 s.

| Check | Result |
| --- | --- |
| Wire baseline (power save on, plain `set_leds` ×100) | 100/100 ok — median 40.8 ms, p90 338 ms, max 1156 ms |
| Stream start | WiFi power save forced off (`max_modem` → `none` reported in start response) |
| 30 fps feed | frames_sent / frames_received ≈ 70% → effective ~20 fps; no errors; visually smooth rotation |
| `event` bypass | every 3 s flash visible on device — no event frame lost while continuous frames were being downsampled |
| Stop | power save restored (post-stop round-trips returned to the 700–900 ms power-save pattern, confirming the restore path) |
| Manual fallback | plain `set_leds` succeeded immediately after stop (3/3), no switching step |

Conclusion on the acceptance latency gate: the ~40–50 ms per-frame round-trip on a
typical home WLAN caps the effective dispatch rate at ~20 fps, so 30 fps is not
achievable on this transport; the practical recommendation is `max_fps: 20` (and
feeding at the same rate to avoid pointless drops). Recorded in the tool schema
description and both READMEs in a6d1b07. The high-jitter baseline (p90 338 ms with
power save on) also confirms the WiFi power-save handling in this PR is required
for stream use.

PR #339, opened 2026-07-10
Add StackChan sensor readings and NFC scans

maha0525:
## What changed

- Add `self.imu.read` for a BMI270 + BMM150 9-axis snapshot.
- Add `self.environment.read` for one LTR-553ALS-WA ambient-light/proximity snapshot.
- Add `self.nfc.scan` for one body-mounted ST25R3916 ISO 14443A UID scan.
- Expose the firmware tools as parameterless gateway tools: `read_imu`, `read_environment`, and `scan_nfc`.
- Document all three tools and add gateway relay tests.

## Safety and scope

The CoreS3 and body peripherals share safety-critical internal I2C buses, so no generic internal-bus read/write tool is exposed.

- IMU returns physical units plus raw samples and data-ready metadata.
- LTR-553 returns its two ambient-light ADC channels and proximity ADC count with ready/valid/saturation flags; it does not start a monitoring stream.
- NFC enables its RF field only for the explicit scan and turns it off afterward. It returns UID / ATQA / SAK only: no tag-memory read/write, authentication, or emulation. UID bytes are not written to firmware logs.

## Validation

- `pytest tests/test_stdio_server.py -q` — 93 passed
- `ruff check stackchan_mcp tests/test_stdio_server.py` — passed
- Previous LTR-553 head commit completed the repository CI, including the fixed `espressif/idf:v5.5.2` firmware build.
- The same CI is running for the NFC head commit.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,7 @@ jobs:
         run: |
           set -euo pipefail
           test -f firmware/build/merged-binary.bin
+          test -f firmware/build/xiaozhi.bin
           test -f firmware/releases/v2.2.6_stackchan.zip
 
       - name: Upload firmware artifacts
@@ -46,6 +47,7 @@ jobs:
           retention-days: 7
           path: |
             firmware/build/merged-binary.bin
+            firmware/build/xiaozhi.bin
             firmware/releases/v2.2.6_stackchan.zip
 
   firmware-host-test:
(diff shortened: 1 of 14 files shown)


kisaragi-mochi:
Heads-up while this is in draft: #355 (currently open for review) also adds a `self.imu.read` device tool on the firmware side (BMI270 accelerometer + a sampling-window statistics variant), so the two branches will overlap on the tool name and on `stackchan.cc`. Nothing you need to do right now — just flagging it early so it doesn't surprise you at rebase time. If both land, the final API shape can be coordinated on this end (worst case one of the tools gets a different name), so please don't let this discourage the draft.

Issue #303, opened 2026-06-18
LCD subtitle overlay — bottom-band single-line scroll synced to say()


kisaragi-mochi:
Parent: #302

## Background

In high-volume environments (= rehearsal studio, outdoor music venues) the device's voice output via `say()` is inaudible to humans. A visual fallback that surfaces the spoken text on the LCD itself would close this gap and also help users who are partially hearing-impaired.

## Goal

Display the active `say()` text as a horizontally scrolling single-line subtitle in the bottom ~1/4 of the LCD, leaving the avatar render area intact. The feature is opt-in via a `say()` argument and does not change existing call sites.

### Layout

- Render an lv_obj subtitle panel aligned at `LV_ALIGN_BOTTOM_MID` with a small bottom margin (~6 px)
- Panel width: ~316 px (≈ LCD width − 4 px), height: a single text-line (~22 px including padding)
- The panel overlays the avatar but does not occlude the eye / mouth regions; `lv_obj_move_foreground()` after creation so it always renders above the avatar layer

### Text handling

- **Single line only.** Multi-line layout is rejected because the added vertical extent forces longer on-screen dwell time and desyncs the subtitle from the spoken phrase
- **Scroll** the text horizontally when it exceeds the panel width. Two options for scroll speed:
  - **(A) sync to TTS duration** — query the speech duration from the gateway side and pace the scroll so the tail clears the panel as the audio ends (preferred if technically feasible)
  - **(B) fixed pace** — scroll at a constant pixels-per-second rate derived from text length (fallback if duration is unavailable)
- **Hide trigger:** when the speech completes (= TTS playback end signal), hide the panel. If the speech is interrupted, hide immediately. Safety timeout (~30 s) to prevent stuck subtitles

### API surface

Extend the existing `say(text, ...)` MCP tool with an opt-in argument:

```python
say(text="Hello", subtitle=True)
```

When `subtitle=True`, the gateway pushes the text to the firmware as part of the say request. Firmware-side: if the subtitle protocol is present in the request, populate and show the subtitle panel; otherwise no change.

### Emoji handling

- Strip face emojis (= the set landed in #294) from the subtitle text by default. The avatar already shows the emoji-driven expression visually; duplicating the emoji as text is redundant
- Non-face emojis (= miscellaneous) pass through unchanged

### Coexistence with the avatar

[...]

PR #354, opened 2026-07-16
feat(stackchan): show_caption subtitle overlay (firmware + gateway relay)

marikagura:
## Summary

Adds an on-screen subtitle path for silent operation, where speech is shown instead of spoken:

- **Firmware**: a `self.display.show_caption` device tool — a translucent bottom subtitle band over the avatar: small white text, at most two lines with ellipsis truncation, fading out automatically after a configurable duration. Font size (14/16/20 px Montserrat, ASCII-only) and band opacity are per-call arguments so the look can be tuned without reflashing.
- **Gateway**: a `show_caption` MCP tool relaying to the device tool, plus list/relay tests.

### Code-level

- [x] gateway: `uv run pytest` (722 passed, 5 skipped)
- [ ] firmware: `python ./scripts/release.py stackchan` — not run locally, CI covers the build

### Hardware

- [x] Device boots without crash
- [x] Existing MCP tools still work where affected: `move_head`, `take_photo`, `set_volume`, `get_head_angles`
- [x] Existing touch/servo behavior still works where affected: tap, stroke, wobble
- [x] New behavior verified on real hardware: in daily service on a CoreS3 stack-chan since 2026-07-09; coexists with lip-sync and avatar rendering, band fades as configured, truncation behaves on two-line overflow

## Breaking changes

None (new tools only).

## Related issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,22 @@ documented-only.
 
 ## [Unreleased]
 
+### Gateway
+
+- Added a `show_caption` tool that relays to the new
+  `self.display.show_caption` device tool, for showing an utterance as an
+  on-screen subtitle in silent operation.
+
+### Firmware
+
+- Added a `self.display.show_caption` device tool: a translucent bottom
+  subtitle band (small white text over the avatar, at most two lines with
+  ellipsis truncation) that fades out automatically after a configurable
+  duration. Font size (14/16/20 px Montserrat, ASCII-only) and band
+  opacity are per-call arguments so the look can be tuned without
+  reflashing. Intended for silent-mode operation where speech is shown
+  instead of spoken.
+
 ## [0.17.0] - 2026-07-12
 
 ### Gateway
--- a/firmware/main/boards/stackchan/config.json
+++ b/firmware/main/boards/stackchan/config.json
@@ -8,7 +8,9 @@
                 "CONFIG_CAMERA_GC0308=y",
                 "CONFIG_CAMERA_GC0308_AUTO_DETECT_DVP_INTERFACE_SENSOR=y",
                 "CONFIG_CAMERA_GC0308_DVP_YUV422_320X240_20FPS=y",
-                "CONFIG_STACKCHAN_SERVO_FEETECH=y"
+                "CONFIG_STACKCHAN_SERVO_FEETECH=y",
+                "CONFIG_LV_FONT_MONTSERRAT_16=y",
+                "CONFIG_LV_FONT_MONTSERRAT_20=y"
             ]
         }
     ]
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -61,6 +61,11 @@ static inline bool ServoWritePosOk(int r) { return r > 0; }
 
 #define TAG "StackChanBoard"
 
+// Caption overlay text font — reuse the board's builtin puhui face
+// (BUILTIN_TEXT_FONT for stackchan in main/CMakeLists.txt) so the caption
+// adds zero flash cost. The basic subset covers everyday CJK.
+LV_FONT_DECLARE(font_puhui_basic_20_4);
+
 class Pmic : public Axp2101 {
 public:
     // Power Init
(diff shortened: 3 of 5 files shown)


kisaragi-mochi (changes requested):
Thank you — an on-screen text path for silent operation fills a real gap, and two things stood out positively on review: the UTF-8 truncation is codepoint-safe (no mid-sequence cuts, CJK counted double), and the CJK-renders-blank limitation is honestly documented in the tool descriptions. Three corrections before this lands, all confirmed against the code and the vendored LVGL 9.5.0 build on this end:

1. **Enforce the two-line contract for newline / wide input** (`TruncateCaptionUtf8` + band sizing). The budget is a character-count heuristic that treats `\n` as an ordinary width-1 character, and the label uses `LV_LABEL_LONG_WRAP` with an unclamped `LV_SIZE_CONTENT` band — so `"A\nB\nC"` renders as three lines and the band grows past the documented two-line maximum. Runs of wide glyphs (`MMMM…`) can likewise exceed the pixel width while staying under the character budget. Either measure with LVGL text APIs or clamp the label/band height to two text lines.

2. **The U+2026 ellipsis has no glyph in the ASCII Montserrat builds** (truncation append). We checked the vendored `lv_font_montserrat_16.c` cmap: U+2026 is absent, and with `LV_USE_FONT_PLACEHOLDER=n` a missing glyph draws nothing with zero advance — truncated text just stops with no visible ellipsis. Appending ASCII `"..."` instead is the simple fix.

3. **`font_size` snapping doesn't match the documented "nearest of 14/16/20"** (`CaptionFontForSize`). The bucket thresholds send 15/17/18/19 all to 16 px, but 19 is nearer to 20 (distance 1 vs 3). A min-distance selection would match the contract stated in the tool description and the gateway schema.

One roadmap note, not a blocker: #303 tracks a planned `say()`-synced subtitle (single-line scroll, TTS-timed hide, and eventually CJK font coverage). This PR's standalone caption is complementary rather than overlapping, and keeping the band/label creation reasonably reusable would be welcome for that follow-up — no action needed in this PR beyond awareness.

After the fixes land we'd like to do a real-device pass on this end (band over the avatar, fade behavior) before merging. No rush.

marikagura:
All three fixed: control whitespace is normalized and the label is clamped to two text lines of the active font; truncation appends ASCII "..."; font_size snaps by minimum distance so 19 selects 20. Verified with a full firmware build. Ready for your on-device pass.

kisaragi-mochi (changes requested):
Thank you for the follow-up — the two-line clamp, the ASCII ellipsis, and the min-distance font snapping all landed as described. The `CHANGELOG.md` conflict was ours (main moved several times while this was open); it has been resolved on this end by merging current `main` in, so nothing is needed from you there.

We flashed this branch and exercised the caption on a CoreS3 device. Two things to resolve before merging, both confirmed on hardware.

**1. The gateway schema omits the numeric ranges the firmware enforces.** `duration_ms`, `font_size` and `bg_opa` state their ranges in the description text, but carry no `minimum` / `maximum` in the JSON Schema, while the firmware's `Property` validation rejects out-of-range values. Confirmed against the device:

| argument | device response |
|---|---|
| `duration_ms: 500` | `{"error": "Value is below minimum allowed: 1000"}` |
| `bg_opa: 200` | `{"error": "Value exceeds maximum allowed: 100"}` |
| `font_size: 99` | `{"error": "Value exceeds maximum allowed: 20"}` |

An MCP client reading the schema has no machine-readable signal that these bounds exist, so a schema-conforming call fails at the device. This matters more than usual here because the primary caller is an LLM working from the schema. Mirroring the firmware bounds as `minimum` / `maximum` would close it.

**2. Wide-glyph input still overflows the two-line contract, and the response misreports it.** `TruncateCaptionUtf8` budgets by character count, so a string of wide glyphs fits the budget while exceeding two rendered lines; the height clamp then drops the third line outright, with no ellipsis appended. Confirmed on device with `"WWWWW WWWWW ..."` (60 characters): only two lines of `W` were displayed, the remainder was cut with no ellipsis, and the call returned `"truncated": false`.

The clamp does hold the two-line contract, which was the primary concern from the previous round — so the band never grows. What remains is that the ellipsis and the `truncated` flag both describe a truncation that the caller cannot see happening. Measuring with the LVGL text APIs to decide the cut point, or at minimum deriving `truncated` from what was actually rendered, would make the response match the display.

For reference, normal-width input behaves correctly: a long English sentence rendered two lines with a trailing `...` and returned `"truncated": true`.

Once these are addressed we will do another device pass and merge. Thank you again for the caption work — the band reads well over the avatar, and the fade timing feels right in practice.

PR #355, opened 2026-07-16
feat(stackchan): expose the BMI270 accelerometer via self.imu.read + imu_read

marikagura:
## Summary

The CoreS3's built-in BMI270 accelerometer was previously untouched by the firmware. This exposes it end to end:

- **Firmware**: a `self.imu.read` device tool returning the latest 3-axis sample in milli-g plus statistics over a 3.2 s / 20 Hz sampling window — per-axis mean and variance, and the orientation-independent variance of the acceleration magnitude, intended for sustained-shake detection by a polling host.
- **Gateway**: an `imu_read` MCP tool (no arguments) relaying to it, plus list/relay tests.

### Code-level

- [x] gateway: `uv run pytest` (722 passed, 5 skipped)
- [ ] firmware: `python ./scripts/release.py stackchan` — not run locally, CI covers the build

### Hardware

- [x] Device boots without crash
- [x] Existing MCP tools still work where affected: `move_head`, `take_photo`, `set_volume`, `get_head_angles`
- [x] Existing touch/servo behavior still works where affected: tap, stroke, wobble
- [x] New behavior verified on real hardware: in daily service since 2026-07-10; we poll it to pause head-tracking during sustained motion (e.g. the unit being carried). No interference with servo or touch paths observed.

## Breaking changes

None.

## Related issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,20 @@ documented-only.
 
 ## [Unreleased]
 
+### Gateway
+
+- Added an `imu_read` tool relaying to the new `self.imu.read` device tool
+  for polling the built-in accelerometer and its recent-window motion
+  statistics.
+
+### Firmware
+
+- Added a `self.imu.read` device tool exposing the CoreS3's built-in BMI270
+  accelerometer: latest 3-axis sample in milli-g plus statistics over a
+  3.2 s / 20 Hz sampling window (per-axis mean and variance, and the
+  orientation-independent variance of the acceleration magnitude, intended
+  for sustained-shake detection by a polling host).
+
 ## [0.17.0] - 2026-07-12
 
 ### Gateway
--- a/firmware/main/boards/stackchan/stackchan.cc
+++ b/firmware/main/boards/stackchan/stackchan.cc
@@ -44,15 +44,34 @@ static inline bool ServoWritePosOk(int r) { return r > 0; }
 #include <esp_lcd_ili9341.h>
 #include <esp_timer.h>
 #include <esp_random.h>
+#include <esp_rom_sys.h>
 #include <freertos/FreeRTOS.h>
 #include <freertos/task.h>
 #include <freertos/semphr.h>
 #include "esp_video.h"
 #include <cJSON.h>
 #include <lvgl.h>
+
+// BMI270 IMU (built into the M5Stack CoreS3, internal I2C bus). The Bosch
+// core driver and its 8 KB config blob come from the espressif/bmi270_sensor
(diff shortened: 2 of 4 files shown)


kisaragi-mochi (changes requested):
Thank you for the patience on this one — and apologies for the conflict, that
was on our side. We have merged `main` into the branch here and kept the
`[Unreleased]` entries from both sides, so there is nothing to do on your end
for that.

The design reads well: the fixed ring with a periodic sampler keeps the read
path off the I2C bus, the back-off after repeated failures protects the shared
bus, and magnitude variance is the right orientation-independent signal to
expose.

Two changes before merge.

**1. The variance is dominated by rounding error at rest**

```c
int64_t mean_a = sum[a] / n;
int64_t var_a  = sum_sq[a] / n - mean_a * mean_a;
```

Both terms are truncated by integer division *before* the subtraction, so the
rounding does not cancel. With gravity on an axis the mean is around 1000 mg,
and dropping the fractional part shifts `mean^2` by up to ~2000 mg². The true
variance at rest is a few mg², so the result is dominated by that error and
lands anywhere in 0..~2000 mg² depending on the fraction.

For the two samples `[1000, 1001]` the population variance is 0.25 mg², while
this computes `1001000 - 1000000 = 1000` mg².

This breaks the contract the tool description advertises — "near 0 at rest
regardless of orientation" — which is exactly what a host calibrates its
threshold against. Transit vibration is large enough (10^4 mg² and up) that a
high threshold still works in practice, but a host that trusts the documented
"near 0" will see sustained false positives on a desk.
`magnitude_variance_mg2`, the headline signal, uses the same expression.

Computing in floating point is enough:

```c
double mean_a = (double)sum[a] / n;
double var_a  = (double)sum_sq[a] / n - mean_a * mean_a;
```

Staying integral, `(n * sum_sq - sum * sum) / (n * n)` evaluated before any
division also works. A short test over a known series (constant, ±1 mg, values
straddling zero) would pin it down.

**2. `available` stays true when the sampler task fails to start**

In `InitializeImu()`, a failed `xTaskCreate` logs and clears `imu_task_`, but
`available` is still derived from `bmi270_.ok()` alone. The tool then reports
`available: true` with `error: "no samples yet"` indefinitely, which reads as
"not warmed up" rather than "not running". Deriving `available` from both the
sensor init and the sampler being alive would make that state honest.

**Out of scope**

The response has no notion of sample age, so a stalled sensor keeps returning
the previous window as a current measurement. That changes the response
contract, so we opened #369 for it rather than growing this PR.

PR #356, opened 2026-07-16
feat(gateway): command-queue watchdog — head timeout + saturation shedding

marikagura:
## Summary

Two failure modes of the HTTP daemon's command queue, addressed with a watchdog:

- **Hung head-of-queue**: a dispatch that never completes used to stall every queued command behind it. It is now force-dequeued after a per-tool timeout (default 30 s; `say`/`listen`/`load_avatar_set` get longer budgets).
- **Saturation lockout**: a full queue used to reject every newcomer uniformly, so a burst of cosmetic LED frames could starve one-shot commands. A saturated queue now sheds the oldest queued *droppable* item (cosmetic LED-class traffic where the next frame supersedes the dropped one); one-shot commands (`say`, `set_avatar`, `take_photo`, `move_head`, …) are never shed.

Interventions are logged and appended to the stackchan JSONL event log (`event_type="queue"`), following the notify configuration's JSONL path when one is set (second commit) so queue events land next to the device events they interleave with.

### Code-level

- [x] gateway: `uv run pytest` (730 passed, 5 skipped — includes a dedicated `test_queue_watchdog.py`)
- [x] Not applicable / not run: firmware build (gateway-only change)

### Hardware

- [x] Not applicable / hardware not available: gateway-only. Running in our deployment since 2026-07-10 — recovered several real hangs (device mid-reflash, Wi-Fi drop during a long TTS dispatch) without manual gateway restarts.

## Breaking changes

None; watchdog defaults preserve existing behaviour for fast tools.

## Related issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,18 @@ documented-only.
 
 ## [Unreleased]
 
+### Gateway
+
+- Added a command-queue watchdog to the HTTP daemon: a hung head-of-queue
+  dispatch is now force-dequeued after a per-tool timeout (default 30 s;
+  `say`/`listen`/`load_avatar_set` get longer budgets) instead of stalling
+  every queued command behind it, and a saturated queue sheds the oldest
+  cosmetic LED-class command to admit new commands (`say`, `set_avatar`,
+  `take_photo`, `move_head` and other one-shot commands are never
+  dropped). Interventions are logged and appended to the stackchan JSONL
+  event log (`event_type="queue"`), following the notify configuration's
+  JSONL path when one is set.
+
 ## [0.17.0] - 2026-07-12
 
 ### Gateway
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,13 +34,16 @@ documented-only.
 
 - Added a command-queue watchdog to the HTTP daemon: a hung head-of-queue
   dispatch is now force-dequeued after a per-tool timeout (default 30 s;
-  `say`/`listen`/`load_avatar_set` get longer budgets) instead of stalling
-  every queued command behind it, and a saturated queue sheds the oldest
-  cosmetic LED-class command to admit new commands (`say`, `set_avatar`,
-  `take_photo`, `move_head` and other one-shot commands are never
-  dropped). Interventions are logged and appended to the stackchan JSONL
-  event log (`event_type="queue"`), following the notify configuration's
-  JSONL path when one is set.
+  `say`/`listen`/`load_avatar_set` get longer budgets, and a
+  `load_avatar_set` call with an explicit `timeout` argument gets a
+  budget derived from that value so schema-legal long fetches are never
+  cut short) instead of stalling every queued command behind it, and a
+  saturated queue sheds the oldest cosmetic LED-class command (Port B
+  and Port C ws2812 frames alike) to admit new commands (`say`,
+  `set_avatar`, `take_photo`, `move_head` and other one-shot commands
+  are never dropped). Interventions are logged and appended to the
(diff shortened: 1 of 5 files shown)


kisaragi-mochi (changes requested):
Thank you — the watchdog closes a real gap (a hung head-of-queue taking out everything behind it is a failure mode we care about), and the dedicated test module is appreciated. Three points to align before this lands, all confirmed against the code on this end:

1. **Honor `load_avatar_set`'s public `timeout` range** (`queue.py` line 72). The per-tool budget is fixed at 180 s while the tool's schema accepts a per-call `timeout` up to 300 s, so a legitimate `timeout=181-300` call gets force-cancelled mid-transfer. The cancellation is also masked: `send_avatar_set_fetch` catches `CancelledError` and returns `{"ok": false, "error": "superseded"}`, so the outer `asyncio.wait_for` returns that value normally and the `HeadTimeout` branch in `run_dispatcher` never fires for this path — the caller sees a misleading `superseded` instead of a head-timeout. (The current head-timeout tests use bare `Event().wait()` stubs, which don't exercise this masking.) A dynamic budget derived from the call's own timeout argument plus margin would keep the watchdog consistent with the public schema.

2. **Include Port C LED commands in the droppable set** (`queue.py` lines 82-93). `DROPPABLE_TOOLS` lists the Port B WS2812 commands but not the `port_c_ws2812_*` equivalents, which are first-class tools with the same next-frame-supersedes semantics. As written, a queue saturated by Port C traffic has no evictable items, so one-shot commands like `say` still get `QueueFull` — the exact starvation this PR sets out to fix.

3. **Respect `jsonl_enabled` for queue events** (`cli.py` line 887). The queue is handed `notify_config.jsonl_path` unconditionally, and `_record_queue_event` → `log_event` appends without checking the flag, so a deployment with JSONL logging disabled still gets the file created on the first drop or timeout. The device-event call site gates on `config.jsonl_enabled`; the queue wiring should follow the same pattern.

No rush on any of this — happy to re-review when updated.

marikagura:
All three points addressed: the dispatch budget now follows load_avatar_set's own timeout argument (clamped to the schema maximum plus margin, with a regression test mirroring the absorbing fetch path), the Port C ws2812 frame commands joined the droppable set, and queue events honor jsonl_enabled. Full gateway suite passes.

kisaragi-mochi (changes requested):
Thank you for the follow-up — the dynamic budget, the Port C additions, and the `jsonl_enabled` gate all check out on this end, and the regression test mirroring the absorbing fetch path is a welcome touch. The conflict on `CHANGELOG.md` was ours (main moved several times while this was open); it has been resolved on this end by merging current `main` in, so nothing is needed from you there.

One point surfaced on this round that we would like resolved before merging.

**`say` runs on a fixed 120 s dispatch budget, but the tool's schema places no bound on utterance length.** Synthesis is streamed at real-time pace, so a long paragraph legitimately streams past two minutes — and once it does, the watchdog cancels it and the caller gets `HeadTimeout` partway through the utterance. Without this PR such a call simply runs to completion, so this would be a new ceiling introduced by the watchdog rather than a pre-existing one. Deriving the budget from the utterance itself — estimated audio duration, or text length as a proxy — in the same spirit as the `load_avatar_set` treatment would keep the watchdog from cutting short a call the schema permits.

Separately, and **not** a blocker for this PR: the cancellation-absorbing dispatch path can still let an expiry surface as an ordinary result (no `HeadTimeout`, no JSONL event) when a dispatch stalls before its inner wait begins. The reproduction window is narrow and callers still receive an error, so it is filed as #366 for tracking rather than held against this branch.

PR #357, opened 2026-07-16
feat(gateway): expose ws session_id in get_status for reboot detection

marikagura:
## Summary

`get_status` now reports the WebSocket `session_id` alongside the connection flags. The id changes on every (re)connection, so a polling host can detect a device reboot even when the reconnect lands between two polls and the `connected` flag never reads false (e.g. a firmware reflash).

### Code-level

- [x] gateway: `uv run pytest` (720 passed, 5 skipped)
- [x] Not applicable / not run: firmware build (gateway-only, 4-line change)

### Hardware

- [x] Not applicable / hardware not available: gateway-only.

## Breaking changes

None (additive field).

## Related issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,13 @@ documented-only.
 
 ## [Unreleased]
 
+### Gateway
+
+- `get_status` now reports the WebSocket `session_id` alongside the
+  connection flags. The id changes on every (re)connection, so a polling
+  host can detect a device reboot even when the reconnect lands between
+  polls and `connected` never reads false.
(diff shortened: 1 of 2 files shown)


kisaragi-mochi (approved):
Reviewed. The additive `session_id` field is the right shape for reboot detection — it changes on every (re)connection, so a polling host can catch a reconnect that lands between two polls without any race against the `connected` flag. Code-level review on this end found no issues; CHANGELOG entry is in place and gateway tests pass. Thank you for the focused change.

PR #358, opened 2026-07-16
docs: avatar authoring notes — geometry, sources, fetch window, blink cadence

marikagura:
## Summary

Adds `docs/avatar-authoring-notes.md` — field notes from building custom avatar sets for the `load_avatar_set` pipeline, covering the authoring side the tool docs do not:

- keeping all frames of a set in one geometry (why near-miss variants read as jumps),
- authoring moving parts (eyes/mouths) as edit-grade layered sources instead of cropping from composited art,
- sequencing around the multi-minute set fetch under Wi-Fi power save,
- treating blink cadence as a per-face aesthetic (and where the constants live).

Advice, not requirements — nothing in it is enforced by gateway or firmware.

### Code-level

- [x] Not applicable / not run: docs-only change (no code paths touched)

### Hardware

- [x] Not applicable / hardware not available: docs-only. The observations themselves come from a CoreS3 stack-chan in daily service (geometry/edge pitfalls hit while iterating a 90-frame matrix set; fetch timing measured on-device).

## Breaking changes

None.

## Related issues

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,8 @@ documented-only.
 ### Docs
 
 - Added avatar authoring notes (`docs/avatar-authoring-notes.md`):
-  frame-geometry consistency, edit-grade sources for moving parts, the
-  avatar-set fetch window, and per-face blink cadence.
+  frame-geometry consistency, full-frame exports from layered sources,
+  the avatar-set fetch window, and blink cadence tuning.
 
 ## [0.17.0] - 2026-07-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ documented-only.
 
 ## [Unreleased]
 
+### Docs
+
+- Added avatar authoring notes (`docs/avatar-authoring-notes.md`):
+  frame-geometry consistency, edit-grade sources for moving parts, the
+  avatar-set fetch window, and per-face blink cadence.
+
 ## [0.17.0] - 2026-07-12
 
 ### Gateway
--- a/docs/avatar-authoring-notes.md
+++ b/docs/avatar-authoring-notes.md
@@ -1,0 +1,69 @@
+# Avatar authoring notes
+
+Field notes from building custom avatar sets for the `load_avatar_set`
+pipeline. The wire format is documented on the tool itself (layered mode =
+14 frames, face 6 + eyes 3 + mouth 5, 537,600 bytes; matrix mode = 90
+pre-composited frames, 6 × 3 × 5, 3,456,000 bytes; both raw RGB565).
+These notes cover the authoring side: what makes a set read well on the
+device, and the pitfalls we hit so you do not have to.
+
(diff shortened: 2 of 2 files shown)


kisaragi-mochi (changes requested):
Thank you for writing these up — authoring-side field notes are exactly the layer the tool docs don't cover, and the geometry/source-separation advice clearly comes from real on-device iteration. One round of corrections before this lands: four statements describe behavior differently from what the firmware/gateway actually implement. Since this document's value is that readers can follow it verbatim, I'd like those aligned first. No rush at all.

1. **Layered mode is a full-frame swap, not compositing** (lines 31–33). The firmware never overlays eyes/mouth onto a base face: the renderer picks exactly one RGB565 image (the axis selected by `active_layer_`) and replaces the screen with it, and RGB565 carries no alpha. The "author from layered sources" advice is worth keeping, but please make explicit that every exported frame must be a complete full-frame image — transparent-part exports will render as broken faces on blink/mouth transitions.

2. **Normal commands are not held during a set fetch** (lines 49–51). The fetch runs on a dedicated FreeRTOS task and the WebSocket receive loop keeps processing; only face/mouth/blink updates are quiesced during the window. `say` and other tools go through normally, so the sequencing advice applies to avatar-affecting calls, not to commands in general — as written it steers readers toward timeout tuning they don't need.

3. **The knob for long fetches is `load_avatar_set`'s own `timeout` argument** (lines 53–57). It defaults to 60 s (max 300 s), so a multi-minute fetch returns `device_timeout` before the device confirms unless that per-call timeout is raised to ~180–300 s. Raising timeouts on subsequent commands does not extend the fetch wait, so the procedure as described won't hold.

4. **Blink cadence is a global build-time constant** (lines 67–69). The blink gap constants live in `stackchan.cc` and apply to every avatar set after a reflash; nothing is stored per set. Please frame it as a device-wide build-time setting rather than a per-set bake-time decision.

Happy to re-review whenever you get to it.

marikagura:
All four statements aligned with the implemented behavior: full-frame swap (with the explicit full-frame export rule), fetch-window quiescing scoped to avatar-facing calls, the load_avatar_set timeout argument as the knob for long fetches, and blink cadence as a device-wide build-time setting.

kisaragi-mochi (approved):
Re-reviewed after the alignment commit — all four points now match the implementation: the full-frame swap (with the explicit full-frame export rule), the fetch window scoped to avatar-facing calls, `load_avatar_set`'s own `timeout` argument as the knob for long fetches, and blink cadence as a device-wide build-time setting. These notes cover a layer the tool docs genuinely don't, so thank you for the careful pass.

On the conflict: that one was on our side. `main` moved several times (#348, #351, #352, #353, #357) while this was open, which drifted the `[Unreleased]` section out from under the branch. It has been resolved on this end by merging the current `main` in and keeping both entries — nothing was changed in the notes themselves, and nothing is needed from you.

Approving; ready to merge.

PR #360, opened 2026-08-01
feat(firmware): auto charge hysteresis to protect always-plugged battery

mochievi-moe:
## Summary

- Add AXP2101 cell-charge enable/disable using a read-modify-write of register `0x18` bit 1 so unrelated PMIC control bits remain unchanged.
- Add **opt-in**, compile-time configurable automatic charge protection. It is disabled by default; when enabled, charging turns on at 30% or below, turns off at 70% or above, and keeps its current state between thresholds.
- Preserve the pre-feature resting behavior: charging is enabled at boot. If the fuel-gauge sample is unavailable, charging is enabled as a fail-safe.
- Advertise `self.power.set_charge_enabled` and `self.power.get_charge_state` as device MCP tools, including automatic-policy state and thresholds in responses.
- Reuse the existing battery polling path for the approximately 60-second policy check; no new task or timer is introduced.

## Behavior

- Default build (`STACKCHAN_CHARGE_AUTO=0`): automatic protection is off and charging starts enabled, matching the previous firmware behavior.
- Opt-in protection build (`STACKCHAN_CHARGE_AUTO=1`):
  - battery <= 30%: charging enabled;
  - battery >= 70%: charging disabled;
  - battery 31-69%: current charging state is preserved;
  - unreadable/invalid fuel-gauge sample: charging enabled as a fail-safe.
- `self.power.set_charge_enabled` applies immediately but is volatile and resets to the compile-time boot default after reboot. When automatic protection is enabled, the next policy check may override a manual setting.

### CI / build

- [x] Firmware build
- [x] Firmware host test
- [x] Gateway test
- [x] Gateway Windows smoke test
- [x] CHANGELOG check
- [x] Exact PR head `6584d993a1130fe8e00c1123da42cfaa1c416e56` built successfully for StackChan with ESP-IDF v5.5.2 in Docker (2026-08-22).

### Real-device verification

Verified on a CoreS3 over the LAN via mDNS on 2026-08-22:

- [x] Default build booted with `auto_enabled=false` and charging enabled.
- [x] `self.power.set_charge_enabled(false)` succeeded; register `0x18` changed from `10` to `8`, and `get_charge_state` read back `charge_enabled=false`.
- [x] `self.power.set_charge_enabled(true)` succeeded; register `0x18` changed from `8` to `10`, and `get_charge_state` read back `charge_enabled=true` with active charging observed.
- [x] Opt-in automatic-protection build at 99-100% reported `auto_enabled=true`, `last_auto_decision=disabled_high_battery`, and charging disabled.
- [x] Device was restored to the default build and read back `auto_enabled=false`, charging enabled, and register `0x18 = 10`.
- [x] Only `xiaozhi.bin` was flashed at `0x20000`; NVS was not erased.

An unrelated pre-existing mDNS multi-interface/address-aggregation problem observed during testing is tracked in #375.

## Breaking changes

None. The MCP tools and compile-time defines are additive, there is no NVS schema change, and automatic charge protection is disabled by default.

## Related issues

- #362 — follow-up runtime protection toggle (requires rebasing onto the current #360 head)
- #367 — pre-existing AXP2101 I2C error handling
- #375 — pre-existing mDNS address aggregation and container-interface ordering


kisaragi-mochi (changes requested):
Thanks for this — the read-modify-write on `0x18`, the compile-time threshold
validation, and the wrap-safe interval handling are all careful work, and the
on-device check at 78% is useful context.

We reviewed this together with #362 and would like to adjust the design
direction before merging. The core point is which way the policy fails.

**1. Charge first; engage protection once the upper threshold is reached**

`Pmic::Pmic()` currently calls `SetChargeEnabled(false)` before any battery
sample is taken, and `ApplyAutomaticChargePolicy(..., startup=true)` keeps that
OFF state for every sample above `STACKCHAN_CHARGE_ON_BELOW`. A device booted at
50% therefore never charges until it drains to 30%, even while plugged in.

We would prefer the opposite ordering: let the cell charge normally after boot,
and engage the protection policy only once the battery reaches the upper
threshold. That preserves the intent (not holding a plugged-in device at 100%)
while keeping "charging" as the resting state.

**2. Treat an unreadable fuel gauge as "keep charging"**

When `TryGetBatteryLevel()` fails or returns a value above 100, the policy leaves
the charge state unchanged. Combined with the OFF-at-boot ordering above, a unit
whose gauge read fails at startup stays with charging disabled indefinitely, and
an always-plugged device can end up with a drained cell that will not recover on
its own. Please fall back to the AXP2101 reset default (charging enabled)
whenever no valid sample is available.

**3. `STACKCHAN_DEFAULT_CHARGE_ENABLED` should default to 1**

With `STACKCHAN_CHARGE_AUTO=0`, the current default of `0` leaves charging
disabled at every boot. That is neither the AXP2101 factory default nor the
behaviour before this PR — turning the new feature off should restore the
previous behaviour.

**4. Ship the feature disabled by default, as an opt-in**

This changes charging behaviour noticeably, so we would like it off unless a
user explicitly enables it. Together with #362's NVS-backed toggle that gives:
off out of the box, and `self.power.set_charge_protection(true)` to enable it at
runtime without rebuilding. Please set the compile-time default accordingly; the
NVS default in #362 should match.

Please also extend the CHANGELOG entry to describe the startup behaviour and the
opt-in default, not just the 30/70 hysteresis.

**Out of scope — tracked separately**

Review also flagged that `I2cDevice::ReadReg` / `WriteReg` abort the firmware via
`ESP_ERROR_CHECK` on a transient I2C error, so the new MCP tools can restart the
device instead of returning an error result. That is pre-existing behaviour
across the whole `Axp2101` path rather than something this PR introduces, so it
is tracked in #367 instead of being asked for here.

kisaragi-mochi (approved):
The current head addresses the earlier requested changes: charge-enabled boot default, fail-safe charging on an unavailable fuel-gauge sample, opt-in automatic protection, and updated CHANGELOG behavior are all present. CI is green, and real-device verification on CoreS3 confirmed manual charge disable/enable readback plus automatic high-battery shutdown in the opt-in build. Approving the current head.

PR #361, opened 2026-08-02
feat(firmware,gateway): persistent StackChan idle screen-off timeout

mochievi-moe:
## Summary

- Add an independent, NVS-backed idle screen-off timeout for the StackChan CoreS3 panel: default 300 seconds, `0` disables it, valid range `0..86400`.
- Put the ILI9342-family panel through a real sleep-in (`LCD_CMD_SLPIN`) + backlight-off cycle instead of just dimming, and wake it again (`LCD_CMD_SLPOUT` + saved brightness) on touch, voice-session start, or any MCP tool call, via a new common `Board::OnUserActivity()` hook (default no-op for other boards).
- Force a full-screen LVGL invalidate on wake so the avatar redraws cleanly instead of showing a stale/torn frame until an unrelated redraw happens to trigger.
- Advertise `self.screen.set_off_timeout(seconds)` / `self.screen.get_off_timeout()` as device MCP tools (NVS-persisted, applied immediately), and wire the gateway's `tool_map` (`set_off_timeout` / `get_off_timeout`) so both are reachable over the existing MCP transport.
- Independent of the existing 60s dim / 300s PowerOff `PowerSaveTimer`; that timer's discharge/sleep_mode/device-state/audio/MCP-transport gates are not appropriate for an always-applicable screen timeout, so this uses its own timer instead of reusing that one.

### Code-level

- [x] Firmware and gateway changes reviewed diff-by-diff for scope; only `firmware/main/{application.cc,mcp_server.cc,boards/common/board.h,boards/stackchan/stackchan.cc}`, `gateway/stackchan_mcp/stdio_server.py`, and `CHANGELOG.md` are touched.
- [x] `git diff --check` clean (no whitespace errors).
- [ ] firmware: `python ./scripts/release.py stackchan` — built successfully in an isolated worktree during development (Docker `espressif/idf:v5.5.2`, exit 0); not rerun against this exact PR branch tip in this environment.
- [ ] gateway: `uv run pytest` / `uv run ruff check .` — not rerun in this environment; the gateway change is a 2-entry `tool_map` addition following the existing `set_brightness` pattern exactly.

### Hardware

- [x] Real-device tested on 二号機 (StackChan CoreS3, MAC `68:ee:8f:d7:40:5c`), 2026-08-02, with the owner present and watching the physical screen:
  - 20-second timeout set via `self.screen.set_off_timeout` → panel reliably went dark on schedule.
  - While the screen was off, the WebSocket/gateway connection stayed up (no disconnect in gateway logs) — screen-off does not affect networking.
  - `set_avatar` woke the screen in one call: backlight and panel came back with a clean, non-torn redraw (owner visually confirmed).
  - Timeout restored to the 300s default afterward and confirmed via `self.screen.get_off_timeout` (`persistent: true`, `persistence: "nvs"`).
  - An earlier same-day disconnect that looked screen-off-related was root-caused separately to a stale NVS `ota_url` entry from an unrelated prior firmware flash, not to this change; that hypothesis was explicitly ruled out before this PR was opened.
- [ ] Existing MCP tools/touch/servo regression pass (`move_head`, `take_photo`, `set_volume`, `get_head_angles`, tap/stroke/wobble) — not re-swept in this session beyond normal use during the E2E above; no regressions observed.

## Breaking changes

None. `self.screen.set_off_timeout` / `self.screen.get_off_timeout` and the gateway `tool_map` entries are additive. The new NVS key (`display` namespace, `off_timeout`) is new and ignored by older firmware. The screen-off timeout defaults to enabled (300s) but only affects panel sleep/backlight — it does not change any existing MCP tool's signature or behavior.

## Related issues

None assigned.


kisaragi-mochi (changes requested):
Thank you for this, and apologies for the conflict — that was on our side. We
have merged `main` into the branch here and kept the `[Unreleased]` entries from
both sides, so there is nothing to do on your end for that.

We flashed this branch and confirmed the timeout itself works on the device: the
panel went dark on schedule at the 300 s default. The structure is good — the
ILI9342C sleep-in/out sequencing with the 120 ms guards, keeping the remaining
delay when activity arrives right after timeout, and the full-screen invalidate
on wake are all handled carefully.

Two changes before merge, and one item we will track separately.

**1. `say` does not wake the panel**

`Board::OnUserActivity()` is reached from `McpServer::DoToolCall()` and from
four call sites in `application.cc` (wake word, `ToggleChatState`,
`StartListening`, `WakeWordInvoke`). The gateway's `say` does not go through any
of them: it sends a `tts.start` message directly, and the handler at
`application.cc:539` transitions to Speaking and starts lip-sync without
signalling activity.

So after the timeout, a `say` plays audio with the panel still dark and the
avatar mouth animating on a screen nobody can see. That is the most visible way
to hit this, and it is the normal path for a host that drives speech rather than
tool calls.

Calling `board.OnUserActivity()` when a valid `tts.start` is scheduled — before
the state change and lip-sync — would cover it. It would also be worth a pass
over the other non-MCP paths that write to the display (emotion updates,
avatar-set application) to decide explicitly which of them should count as
activity.

**2. `OnUserActivity()` does not reset the existing `PowerSaveTimer`**

`StackChanBoard::OnUserActivity()` re-arms the screen timer only. The board also
runs `PowerSaveTimer(-1, 60, 300)`, which dims at 60 s and calls
`pmic_->PowerOff()` at 300 s, and `power_save_timer_->WakeUp()` is currently
reached only from `SetPowerSaveLevel()`.

Once `PowerSaveCheck()` has entered sleep mode it stops re-evaluating
`CanEnterSleepMode()`, so the shutdown deadline keeps advancing regardless of
activity. On battery (`SetEnabled(discharging)`), a user who touches the device
to bring the screen back can still have it power off shortly afterwards.

This predates the PR, but the PR makes "screen goes dark, user wakes it" an
everyday interaction, so it is worth resetting both deadlines from the same
activity. Adding `power_save_timer_->WakeUp()` to `OnUserActivity()` matches what
the other boards in the tree already do from their touch handlers.

**3. Tracked separately: the wake path runs on the shared esp_timer task**

`ExitPanelSleepLocked()` can block for up to 240 ms in `vTaskDelay` (the two
120 ms ILI9342C guards) while holding the display lock, and the timeout callback
itself runs on `ESP_TIMER_TASK`. Touch polling and the power-save tick share that
task, so they stall for the duration of a wake.

The user-visible effect is small — a brief hitch at the moment the screen comes
back — and fixing it properly means moving the sequence onto a dedicated task
with a state machine, which is more than this PR should carry. We will open a
separate issue for it rather than expand the scope here.

kisaragi-mochi (approved):
The two requested behavior fixes are resolved at this head (`4c39a2a0bf9677878c19f2fcf8bf0de4bd10a44f`):

- gateway `tts.start`, emotion updates, and avatar-set fetches now signal `OnUserActivity()`, so those visible/audio interactions wake the panel;
- `StackChanBoard::OnUserActivity()` now resets the existing `PowerSaveTimer` as well as the independent screen-off timer.

The CHANGELOG matches the updated behavior. I reran the gateway suite from this worktree (724 passed, 1 skipped), and Ruff and `git diff --check` pass. The exact head also had a successful firmware Docker build during the earlier fix verification; I did not repeat flashing or physical LCD/NVS testing in this review.

Approved on code behavior. One integration task remains before the owner can use the merge button: GitHub currently reports this PR as conflicting with current `main` in `CHANGELOG.md` and `firmware/main/boards/stackchan/stackchan.cc` after #360 merged. Please resolve that current-main conflict without dropping either the charge-control or screen-timeout behavior.

PR #362, opened 2026-08-02
feat(firmware,gateway): persistent charge protection toggle (depends on #360)

mochievi-moe:
Depends on #360 (this branch includes #360's commit `bbd74cb` as an ancestor; merge #360 first and this PR will reduce to only the `charge_protection` change).

## Summary

- Make the 30%-70% battery-protection hysteresis from #360 persistently toggleable at runtime instead of compile-time-only: adds an NVS-backed `charge_protection` flag (namespace `stackchan`, key `chg_protect`; default `true`/enabled when never saved).
- With protection on (default), firmware preserves the existing hysteresis: charging enabled at 30% or below, disabled at 70% or above, unchanged between thresholds.
- With protection off, that policy no longer stops charging; the AXP2101 charger stays enabled and its own full-charge termination remains in control — useful for topping up to 100% before going out.
- Add `Settings::SetBoolAndCommit` (`firmware/main/settings.cc`/`.h`) for an explicit, immediately-committed NVS write independent of the destructor's batched commit.
- Advertise `self.power.set_charge_protection(enabled)` / `self.power.get_charge_protection()` as device MCP tools; wire the gateway's `tool_map` (`set_charge_protection` / `get_charge_protection`) so both are reachable over the existing MCP transport, matching the `set_brightness` pattern.
- Changes take effect immediately without rebooting; switching protection back on immediately re-evaluates the current battery level.
- Document the new tools and the `charge_protection` NVS mapping in `gateway/README.md`.

### Code-level

- [x] gateway: `uv run pytest` — full gateway suite (725 tests) passes, including the new `set_charge_protection`/`get_charge_protection` coverage in `gateway/tests/test_stdio_server.py`.
- [x] firmware: Docker build (`espressif/idf` toolchain, targeting the existing v2.2.6 release path) succeeds with the `charge_protection` NVS plumbing and MCP tool wiring in `firmware/main/boards/stackchan/stackchan.cc`.
- [ ] Real-device revalidation of the persisted flag across a power cycle was not repeated in this PR beyond the code-level checks above.

## Backward compatibility

Additive only. `charge_protection` defaults to enabled (`true`), which reproduces #360's always-on hysteresis behavior for anyone who hasn't called `set_charge_protection`. No existing MCP tool signature changes. `self.power.set_charge_enabled` / `self.power.get_charge_state` from #360 are untouched.

## Related issues

None assigned. Depends on #360 as noted above.


kisaragi-mochi (changes requested):
Reviewed alongside #360 — please see the design notes there; this branch will
need to follow the same direction (opt-in default) once #360 is updated.

Two additional requests for this PR:

- Make the ON/OFF thresholds runtime-configurable through NVS as well, not just
  the enable flag. Adjusting the band without a rebuild is the main practical
  benefit of moving this to runtime control.
- The CHANGELOG entry here still only describes the #360 hysteresis; please add
  the `charge_protection` toggle and its NVS default.

kisaragi-mochi (changes requested):
Thanks for the follow-up — the runtime threshold persistence, opt-in default, gateway aliases, and updated CHANGELOG now address the previous design requests. I also reran the gateway suite against this worktree: 731 passed, 1 skipped, and Ruff passed.

There is one remaining integration issue before this can merge:

**Please rebuild this branch on top of the current #360 head (`6584d993`).**

This branch still uses the reconstructed `bbd74cb` version of #360 rather than #360's current commit. A merge-tree check between #360's current head and this PR's head (`316dbfd`) produces conflicts in `CHANGELOG.md` and `firmware/main/boards/stackchan/stackchan.cc`, including the PMIC constructor, startup/default behavior, automatic charge policy, and MCP tool section. Therefore, merging #360 first will not currently reduce this PR to only the `charge_protection` changes as the description says.

Please rebase or reconstruct #362 on top of `6584d993`, keep only the NVS toggle/threshold, gateway alias, tests, and documentation changes specific to #362, then rerun the gateway tests and firmware build.

Please also update the PR description after the rebase: it still says protection defaults to `true`/enabled and identifies `bbd74cb` as the #360 ancestor, while the current agreed behavior is opt-in (`false` when unset).

The #362-specific implementation otherwise looks sound. Real-device NVS persistence across a power cycle and physical charge transitions remain to be verified separately.

PR #364, opened 2026-08-08
feat(gateway): add Fish Audio TTS engine

duc795-collab:
## Summary

Adds a `fish-audio` TTS engine backed by the hosted [Fish Audio](https://fish.audio) API.

Fish Audio has strong English voices and user-trainable voice models, which makes it a
practical alternative to VOICEVOX (Japanese-only by default) for English deployments.

The reason it is cheap to add: the API returns 16-bit mono PCM/WAV at a caller-chosen
sample rate, so the engine requests `DEVICE_SAMPLE_RATE` (16 kHz) directly and hands the
PCM to the orchestrator with **no decode**, and in the common case no resample. That keeps
it on the base `[tts]` extra — no `miniaudio` (Irodori) and no `ffmpeg` (Edge TTS). The
`tts-fish-audio` extra is therefore a no-op alias over `[tts]`, provided so users can
declare intent explicitly.

`mp3` and `opus` response formats are deliberately rejected: consuming them would pull in
a decoder purely to undo an encode we never needed.

Configuration is lazy, so an unset key leaves the engine listed but fails synthesis with an
actionable error rather than breaking registration:

| Variable | Default | Notes |
|---|---|---|
| `STACKCHAN_FISH_AUDIO_KEY` | _(required)_ | Sent as `Authorization: Bearer <key>` |
| `STACKCHAN_FISH_AUDIO_MODEL` | _(none)_ | Voice model ID (`reference_id`); omitted when unset |
| `STACKCHAN_FISH_AUDIO_BACKEND` | `s2.1-pro` | Sent as the `model` header |
| `STACKCHAN_FISH_AUDIO_URL` | `https://api.fish.audio/v1/tts` | Endpoint override |

Includes README docs mirroring the existing Irodori subsection, and a CHANGELOG entry.

### Note for reviewers

`gateway/uv.lock` is intentionally **not** regenerated here. Running `uv sync` locally adds
the `tts-fish-audio` extra to the lock but also rewrites unrelated resolution markers (for
example `typing-extensions` and `onnxruntime` lose their `python_full_version < '3.11'`
markers), which looks like a uv-version difference rather than an intended change and could
affect resolution on Python 3.10. CI's `uv sync --frozen --extra tts` is unaffected, and
PyPI installs do not consult the lockfile. Happy to add a re-locked commit if you would
prefer it in this PR.

### Code-level

- [x] gateway: `uv run pytest` — 744 passed, 6 skipped. The 2 `test_ownership.py` cases that
      call `os.sysconf` are Linux-only and cannot run on the Windows dev machine used here;
      they are unrelated to this change and are covered by the Ubuntu CI job.
- [x] gateway: `uv run ruff check .` — All checks passed (ruff 0.15.12, as pinned in `uv.lock`).
- [ ] firmware: `python ./scripts/release.py stackchan`
- [x] Not applicable / not run: firmware build — this is a gateway-only change.

New `tests/test_fish_audio.py` covers the HTTP layer with `httpx.MockTransport` (no network):
request shape, `reference_id` selection and per-call `speaker_name` override, PCM passthrough,
odd-length PCM trimming, WAV decode, resampling when the returned rate differs, rejection of
unsupported formats and empty text, and the 401 / 402 / other error paths. 27 tests, all passing.

### Hardware

- [ ] Not applicable / hardware not available: gateway-only change. TTS runs entirely on the
      gateway; the firmware only plays the Opus frames it is handed, so no firmware behavior
      changes. End-to-end synthesis against the live Fish Audio API has **not** been run here
      (it would consume API credits), so the engine is verified by unit tests plus a runtime
      registration check — `get_registry().names()` returns
      `['edge-tts', 'fish-audio', 'irodori', 'voicevox']`.

## Breaking changes

None. New opt-in engine and a new extra; existing engines, defaults, and the `say` API are
unchanged. VOICEVOX remains the default when `STACKCHAN_TTS_ENGINE` is unset.


kisaragi-mochi (changes requested):
This is a well-put-together PR. The no-decode path is a genuinely good fit for
the device sample rate, the test file pins the request shape and the error codes
without touching the network, and explaining up front why `uv.lock` was left
alone saved a round trip. Two things to change before merge, one doc item we are
taking off your plate, and one finding that belongs elsewhere.

**1. A malformed 200 response is played as audio**

On HTTP 200 the body is accepted as PCM without checking the content type or the
declared format, and an odd-length body is trimmed by one byte and reported as a
success. A captive portal or a proxy that answers 200 with HTML or JSON therefore
becomes noise on the device speaker, and a truncated response becomes a silently
shortened utterance that still reports success. The WAV path does not verify that
the declared frame count was fully read either.

This is specific to this engine rather than a pre-existing gap: VOICEVOX and
Irodori push their responses through a decoder, so a malformed body fails there.
The no-decoder design is the appealing part of this PR, but it also removes the
layer that was implicitly validating the response — so the validation needs to
become explicit here.

We would like the content type and declared format checked, odd-length PCM
rejected rather than trimmed, and the WAV path to confirm the declared frames
were fully read. Tests for a 200 non-audio body, odd-length PCM, and a truncated
RIFF would pin it.

**2. `tts-fish-audio` is missing from `uv.lock`**

`pyproject.toml` gains the extra, but the lockfile has no matching entry under
optional-dependencies / requires-dist / provides-extras. `uv sync --frozen
--extra tts-fish-audio` therefore installs neither httpx nor opuslib, and the
failure only surfaces at synthesis or encode time. CI's `--extra tts` is indeed
unaffected, but the dedicated extra is the one the README tells users to install.

Your offer to add a re-locked commit is welcome — please go ahead. If the
unrelated marker churn on `typing-extensions` / `onnxruntime` reappears, say so
and we will sort out the uv-version difference on our end rather than have you
fight it.

**3. Documentation — the Japanese README is on us**

The repository keeps `README.md` and `README.ja.md` in sync, and `README.ja.md`
already documents the VOICEVOX and Irodori engines. Writing the Japanese section
is not something to hand to a contributor, so we will prepare it. Nothing to do
on your side for that.

One addition we would like in the English section: a sentence stating that the
text passed to `say` is sent to a third-party hosted service, and that users
should review Fish Audio's terms and pricing. The current wording says "hosted"
but does not spell out that utterance text leaves the machine. Either add it
yourself, or leave it and we will fold it in together with the Japanese section —
whichever you prefer.

**Out of scope — tracked separately**

Review also flagged that the numeric httpx timeout bounds each chunk rather than
the whole synthesis call, so a slow-drip response can hold the `say` pipeline
open well past the configured value. VOICEVOX and Irodori share that pattern, so
it is tracked in #368 rather than asked for here.

duc795-collab:
Pushed 6246d8d addressing points 1–3.

**1. Malformed 200 responses**

Validation is now explicit in the engine rather than inherited from a decoder, since not carrying one is what removed that layer in the first place. Four rejection paths:

- Content type is not audio. `application/octet-stream` and `binary/octet-stream` are accepted alongside `audio/*`, because headerless `pcm` has no content type of its own and proxies relabel it.
- The body is the other supported format. A WAV returned when `pcm` was requested played its RIFF header as a burst of noise before the audio; the reverse fed a headerless body to the `wave` reader.
- A WAV shorter than its own header declares. `wave.readframes` returns what it has rather than raising, so a clipped download became a clipped utterance reported as success.
- An odd-length `pcm` body. This was previously trimmed by one byte to realign the framing, which returned audio quietly missing its tail — it is now reported as truncation.

The frame-count check sits in the engine rather than in `tts/audio_utils.wav_to_pcm16_mono`, since VOICEVOX shares that helper and, as you noted, the gap is specific to the undecoded path. Happy to move it if you would rather the helper be strict for everyone.

`test_odd_length_pcm_is_trimmed` becomes `test_odd_length_pcm_is_rejected` — the behaviour it pinned is the one being changed. Four tests cover the new rejections (non-audio 200 as HTML and as JSON, truncated RIFF, format mismatch). All five fail against the previous engine.

**2. `uv.lock`**

Added. `uv sync --frozen --extra tts-fish-audio` previously installed neither `httpx` nor `opuslib`; it now pulls both, and `uv lock --check` is clean.

To answer the question you raised: **the marker churn does reappear.** A full `uv lock` on uv 0.12.3 drops `python_full_version` markers across `typing-extensions` and the whole `onnxruntime` tree — `flatbuffers`, `numpy`, `packaging`, `protobuf`, `sympy`, `mpmath` — which is unrelated to this extra. I have therefore hand-applied only the three `tts-fish-audio` entries (`optional-dependencies`, `requires-dist`, `provides-extras`), taken verbatim from what uv generated, so the lockfile diff is limited to this feature. Taking you up on the offer to sort out the uv-version difference on your end.

**3. Documentation**

Added a sentence to the English section stating that text passed to `say` leaves the machine for a third-party service, with a pointer to Fish Audio's terms and pricing and a suggestion to prefer a local engine for sensitive utterances. Also documented the new rejection behaviour. Leaving `README.ja.md` to you as offered.

**Not addressed:** the httpx timeout bounding each chunk rather than the whole call — following your lead in tracking that as #368.

One note on verification: the gateway suite passes (753 passed, 1 skipped), but two pre-existing `test_ownership.py` failures show up in my environment because `os.sysconf` is POSIX-only and I am on Windows. They reproduce on the branch point without my changes and are unrelated. CI on Linux should be unaffected.

kisaragi-mochi (changes requested):
Thanks for the detailed follow-up. The previously requested odd-PCM, RIFF/format, truncated-WAV, lockfile, and English privacy/terms changes are all present at `6246d8d1a8cbac8073a93d69207c24108d594fdf`. I also checked the current Fish Audio OpenAPI directly: `/v1/tts`, Bearer auth, PCM/WAV at 16 kHz, and the documented `s2.1-pro` model header are valid.

Validation on this head is clean: Fish Audio focused tests pass (31), the full gateway suite passes (755 passed, 1 skipped), `uv lock --check`, Ruff, and `git diff --check` all pass.

There is one remaining malformed-200 hole before merge:

- `_assert_audio_response()` only rejects a non-audio Content-Type when `kind` is non-empty. If a proxy/captive portal returns HTTP 200 with no Content-Type and an even-length HTML/JSON body, the empty `kind` bypasses the check and the body is returned as headerless PCM, so the robot plays it as noise. For example, `_assert_audio_response("", b"<html>", "pcm")` currently succeeds. Please reject a missing/empty Content-Type too, and give successful mock PCM/WAV responses an explicit audio Content-Type so this path is pinned by a test.

The Japanese README synchronization remains the maintainer-owned documentation item from the previous review; I am not asking the contributor to take that on. I did not call the live paid API or repeat robot playback during this review.

PR #365, opened 2026-08-08
chore: add Windows stdio runner script

duc795-collab:
## Summary

Adds `start-stdio.ps1`, a small Windows convenience runner for the stdio MCP server.

It loads `gateway/.env` into the process environment and then starts the server from the
gateway directory. The gateway already calls `load_dotenv()` itself, but exporting the
variables here keeps them visible to `uv` and to any child process it spawns.

The parser handles `KEY=value` and `KEY="value"` lines, strips one layer of matching quotes,
and skips comments, blanks, and empty values so an unset placeholder in `.env` never shadows
a real default. It fails fast with an actionable message when `gateway/.env` is missing.

Windows-only convenience; on Linux the existing bash scripts apply. No secrets are embedded —
everything is read from `gateway/.env`, which is gitignored.

Split out of the Fish Audio engine PR, since it is unrelated tooling and CONTRIBUTING asks
for focused PRs.

### Code-level

- [ ] gateway: `uv run pytest`
- [ ] gateway: `uv run ruff check .`
- [ ] firmware: `python ./scripts/release.py stackchan`
- [x] Not applicable / not run: this adds a single root-level PowerShell script and touches no
      Python or firmware source. Verified manually by running it — it loads `.env` and starts
      the stdio server.

Note: the CHANGELOG gate in `build.yml` only triggers on `firmware/*` or `gateway/*` paths, so
this root-level script correctly does not require a CHANGELOG entry.

### Hardware

- [x] Not applicable / hardware not available: developer tooling only, no device interaction.

## Breaking changes

None. New standalone file; nothing existing references or depends on it.


kisaragi-mochi (changes requested):
Thanks for putting this together — Windows setup is a real gap, and the quoting
and empty-value handling in the parser shows care.

Before asking for a rewrite, we would like to check the intended use, because
review surfaced two issues that would require reworking most of the script.

**1. The startup output breaks the stdio protocol**

`Write-Host "loaded $key"` runs before the server starts. For a stdio MCP server
stdout *is* the JSON-RPC channel, so anything emitted ahead of the first message
can make the client fail initialization. Diagnostics for this process need to go
to stderr, or not exist at all on the normal startup path.

**2. Pre-setting the environment can lock in a wrong value**

The parser handles the current `.env.example`, but it is not equivalent to
python-dotenv, which the gateway uses. Lines such as
`STACKCHAN_TOKEN=secret # comment`, escaped quotes, and multi-line values are
read differently. Because `load_dotenv()` does not override variables that are
already set, a value parsed differently here becomes the final value and the
Python side cannot correct it — which then surfaces as an authentication failure
rather than a parse error. `export KEY=value` lines would also produce a key
named `export KEY`.

Two smaller points: a non-zero exit from `uv` is not propagated (a launch failure
returns 0 to the caller), and the description mentions "the existing bash
scripts" for Linux, but the repository does not currently ship gateway startup
scripts in any shell.

**The question**

The documented setup registers the server directly in the client config:

```json
"command": "uv",
"args": ["run", "--directory", "<path>/gateway", "python", "-m", "stackchan_mcp"]
```

with environment variables supplied there, and the gateway calls `load_dotenv()`
on its own. That path works on Windows as well.

Was there a specific situation where that was not enough? If there is a concrete
gap, we would rather close it in the documented flow. Alternatively we can take
this script with the console output removed and the parsing delegated to
python-dotenv, placed under `gateway/`, with a Windows registration example added
to both `README.md` and `README.ja.md`. If it was mainly a local convenience,
keeping it outside the repository may be the simplest outcome.

duc795-collab:
Thank you for your thoughtful feedback. I truely appreciate it.  Please do
keep outside
of repository as this was more for local convenience.

Regards

kisaragi-mochi:
Understood — thanks for confirming, and thanks for taking the time to put this
together in the first place.

Closing this as agreed, with the script kept outside the repository.

The Windows gap itself is still noted on our side: the documented registration
path (`uv run --directory <path>/gateway python -m stackchan_mcp` in the client
config) works on Windows, but the README does not spell that out for Windows
users. We will look at documenting it separately.

Thanks again for the report and the quick follow-up.

PR #371, opened 2026-08-15
feat(gateway): multi-device routing for say/listen/load_avatar_set

mochievi-moe:
## Summary

Replace the gateway's single ESP32 connection slot with a registry keyed
by the `Device-Id` header, so two or more StackChan units can stay
connected to one gateway process at the same time.

- Adds an optional `device_id` argument to `say`, `listen`,
  `load_avatar_set`, and every generic hardware-relay tool that goes
  through `tool_map`.
- Adds two new tools: `list_devices` (enumerate connected units and the
  current default) and `set_default_device` (change which device
  receives calls that omit `device_id`).
- Moves the TTS/STT audio-path lock and the per-hardware-lane locks from
  `ESP32Manager` onto each `ESP32Connection`, so concurrent devices no
  longer share (and cannot step on) each other's lock state.

**Backward compatibility**: `device_id` is optional everywhere. Omitting
it keeps routing to the default device (the first one that connected),
which matches pre-phase2 single-device behavior exactly — existing MCP
clients that don't know about `device_id` need no changes.

**Hardware**: gateway-only change, no firmware touched. Two-unit
same-gateway connectivity has been exercised manually against real
hardware during development; a fuller two-device end-to-end pass is
still planned but not required for this gateway-only PR per
CONTRIBUTING.md's Hardware Test Requirement (marked N/A below).

### Code-level

- [x] gateway: `uv run pytest` — 740 passed, 5 skipped
- [x] gateway: `uv run ruff check .` — all checks passed
- [ ] firmware: not applicable (no firmware changes)

20 new tests added (`test_multidevice_routing.py`,
`test_multidevice_voice_routing.py`) covering: two devices coexisting
without kicking each other, explicit-device_id routing, default-device
fallback when device_id is omitted, unknown-device_id error handling
(soft error, not a raise), reconnect-replaces-old-connection, and
per-device lock isolation — across the connection registry, `say`,
`listen`, and `load_avatar_set` paths.

Also updated pre-existing test doubles in `test_stdio_server.py`,
`test_http_server.py`, and `test_follow_led_stream.py` whose
`call_tool()` stand-ins did not accept the new `device_id` keyword, and
one `BYPASS_TOOLS` exact-set assertion to include the two new registry
tools.

### Hardware

- [ ] Not applicable / hardware not available: gateway-only change,
      firmware untouched; see Summary for the manual two-unit
      connectivity note.

## Breaking changes

None. `device_id` is additive and optional on every tool; the two new
tools (`list_devices`, `set_default_device`) are additive. No NVS
schema or build flag changes.

## Related issues

N/A


kisaragi-mochi (changes requested):
Thanks for taking on the multi-device routing foundation. The registry/default-device behavior and per-device TTS routing are headed in a good direction, and this exact head (`869ff0ce06dcfadef3c6348de0163165cff13ef0`) passes the full gateway suite (745 passed, 1 skipped), Ruff, and `git diff --check`.

I found two user-visible gaps that need addressing before merge:

1. **Please advertise `device_id` in every tool schema that supports per-call routing.** The dispatcher currently pops and forwards an undeclared extra field, but the published schemas for `say`, `listen`, `load_avatar_set`, and generic relay tools such as `move_head` and `set_volume` do not contain `device_id`. Normal MCP clients build their available arguments from those schemas, so users cannot discover or reliably issue the per-call routing that the PR and CHANGELOG promise. Only `set_default_device` currently declares it. A schema-level test over all intended routed tools would keep the contract from drifting.

2. **Please reconcile per-device listen locks with the still-global recording slot.** `audio_stream.py` still has one module-level `_recording_session_id` / frame buffer / owner. Device A can acquire its own listen lock and start recording, then device B acquires a different lock but fails at the global `is_recording()` guard with “recording slot is already held”; device-driven capture on B is also ignored while A owns that slot. This contradicts the PR's claim that concurrent devices no longer share STT lock state. Either make recording state truly per-device, or retain an honest gateway-global capture lock/queue and document that limitation so a second request waits instead of unexpectedly failing.

The new routing tests pass, but they currently test the internal forwarding helpers rather than the published MCP schemas or two simultaneous listen windows, which is why these paths remain uncovered. I did not repeat the real two-device hardware/WebSocket exercise during this review.

PR #372, opened 2026-08-17
feat(tts): add ElevenLabs engine

tsuru0805:
Adds ElevenLabs as a third TTS engine, following the Irodori engine's thin-client shape: POST to `/v1/text-to-speech/{voice_id}`, decode the MP3 reply to 16 kHz mono PCM, hand it to the orchestrator.

- Voices map through `STACKCHAN_ELEVEN_VOICE_<NAME>` env vars; raw ElevenLabs voice IDs pass through verbatim; `STACKCHAN_ELEVEN_DEFAULT_SPEAKER` (or the sole configured voice) fills in when a call omits the speaker.
- API key from `ELEVENLABS_API_KEY` / `STACKCHAN_ELEVENLABS_KEY`, environment-only.
- New `tts-elevenlabs` extra (`[tts]` + `miniaudio`, mirroring `tts-irodori`); the shared MP3 decode boundary moved to `audio_utils.py` so its install hint can name both extras. `uv.lock` updated.
- Documented in both READMEs alongside the Irodori section.
- 15 tests (httpx.MockTransport, decode boundary monkeypatched — same approach as the Irodori tests). Full suite green (740 passed).

Tested end-to-end on real hardware (CoreS3 StackChan, `eleven_v3`): speech plays through the device speaker, emoji→expression working.


kisaragi-mochi (changes requested):
Thanks for the thorough ElevenLabs integration and hardware verification notes. The main implementation looks sound: I checked the current official API contract, the endpoint/auth header/request body/default MP3 response all line up, the English and Japanese docs are synchronized, and no API key is persisted or logged.

I also verified this exact head (`f080365a4c71530a28e818c7a4c0a4ee68cfb956`):

- focused ElevenLabs + Irodori tests pass;
- full gateway suite passes: 739 passed, 1 skipped;
- Ruff and `git diff --check` pass;
- the shared MP3 decoder move preserves the Irodori path.

One repository-required change is still needed before merge:

- Please add a concise ElevenLabs entry under `CHANGELOG.md` → `[Unreleased]` → `### Gateway`. This PR touches `gateway/`, so `CONTRIBUTING.md` requires that subsection to change and the CHANGELOG CI gate checks it.

I did not repeat the live ElevenLabs credential call or robot playback during review; those remain covered by the real-hardware result reported in the PR description.

kisaragi-mochi (approved):
Thanks for the follow-up. I verified the current head f5bad4c: the requested ElevenLabs entry is now present under CHANGELOG.md > [Unreleased] > Gateway, with no new blocking findings. The focused ElevenLabs tests pass (15), the Irodori/TTS regression set passes (43), the full gateway suite passes (739 passed, 1 skipped), and Ruff plus git diff --check are clean. I did not repeat a live ElevenLabs credential call or robot playback in this follow-up review.

PR #373, opened 2026-08-17
feat(mcp): return take_photo captures as inline ImageContent

tsuru0805:
`take_photo` used to reply with only a text receipt whose JSON points at `image_path` on the gateway host — an MCP client on the other side of the boundary usually can't open that path, so the model never actually saw the frame. This appends the JPEG as an MCP `ImageContent` block (base64) next to the unchanged text receipt.

Guard rails, since `image_path` originates in a device response:

- only string paths resolving inside the capture directory are eligible (realpath prefix check; symlink escape covered by a test)
- JPEG magic required; inline capped at 200 KB (device JPEGs are a few KB)
- every failure mode (wrong type, missing/oversize/empty/non-JPEG file, escape attempt, non-JSON receipt) degrades to the previous text-only reply — a broken image never breaks the tool call

14 new tests incl. dispatch-level; stale `list[TextContent]` annotations in `http_server.py` refreshed. Suite green (738 passed).

Verified end-to-end on real hardware through streamable-http: text receipt + JPEG block arrive together and the client-side model really sees the frame.


kisaragi-mochi (changes requested):
Thanks for fixing the user-visible gap here. I confirmed that the main path works: the low-level stdio handler serializes the unchanged text receipt plus an `image/jpeg` content block, the full gateway suite passes (737 passed, 1 skipped), and Ruff and `git diff --check` are clean.

One repository-required change is still needed before merge:

- Please add this user-visible gateway change under `CHANGELOG.md` → `[Unreleased]` → `### Gateway`. `CONTRIBUTING.md` requires every PR touching `gateway/` to update that subsection, and the CHANGELOG CI gate checks it.

Two non-blocking cleanup suggestions while updating the branch:

- `create_server()`'s registered `call_tool` callback still says `-> list[TextContent]`, although `take_photo` now returns `list[TextContent | ImageContent]`.
- `docs/architecture.md` still describes the old photo flow as path-only; mentioning the inline image block plus retained path/text fallback would keep the architecture docs accurate.

The 200 KB cap and graceful text-only fallback look appropriate for this single-user gateway. I also verified that the PR still points at `c0c5057185aec75679dadfdd864d3b256a198212` before submitting this review.

kisaragi-mochi (approved):
Thanks for the follow-up. I verified the current head dd0e6f9: the required CHANGELOG entry is present, the call_tool return annotation now includes ImageContent, and docs/architecture.md describes the inline JPEG plus retained text/path fallback. No blocking findings remain. The focused inline-photo tests pass (13), the full gateway suite passes (737 passed, 1 skipped), and Ruff plus git diff --check are clean. I did not repeat live hardware capture or a real MCP client display in this follow-up review.

PR #374, opened 2026-08-17
fix(firmware): make head-touch actually responsive

### LINKED ISSUE #2: Touch: Si12T tap detection occasionally drops events
## Symptom

The Si12T head touch sensor (with the asymmetric debounce currently in `firmware/main/boards/stackchan/stackchan.cc`: 200ms press / 400ms release, 100ms polling) reliably detects strokes. Short taps ("ぽん"), however, are sometimes missed when the press window is shorter than ~200ms or when the sensor is recalibrating.

## Possible causes

- 200ms press confirmation may be too long for fast taps
- Si12T's auto-recalibration triggers brief all-false windows mid-touch
- Default sensitivity registers (SEN1–SEN6 at 0x02–0x07) may be too low for short presses

## Approaches to try

1. Tune sensitivity registers SEN1–SEN6 (0x02–0x07) — increase per-zone gain
2. Shorten press confirmation (200ms → 100ms or 50ms) and rely on release debounce to reject jitter
3. Track a "rising-edge candidate" that survives a single all-false sample (bridge over recalibration glitches)

## Acceptance

- Reliable detection of taps as short as 80–100ms
- No new false positives during stroke / no-touch

tsuru0805:
Fixes #2.

On a stock unit, touching through the head shell barely registers (a 20 s continuous stroke produced one hit), quick taps classify as strokes, and the second tap of a quick tap-tap sequence disappears. Five changes to the Si12T path:

1. **Configure the sensitivity registers at init** — the driver never wrote them, so the chip ran power-on default thresholds. TSM12-compatible mapping; CH1 (higher shell-coupling noise floor on my unit: continuous false-fires at the most sensitive 0.50 % step) gets 0.90 %, other channels 0.50 %. Values are read back and logged at init so per-unit variation is diagnosable.
2. **Measure gesture duration between the first samples of the press/release states**, not the debounce-confirmed edges — the 4-sample release confirm inflated every gesture by ~300–400 ms, pushing even the quickest tap past `STROKE_MIN_MS`, so TAP was effectively unreachable.
3. **Split rapid taps instead of merging them** — a re-press after a ≥2-sample gap with a tap-shaped first segment emits that tap immediately and restarts the gesture. 1-sample gaps still bridge strokes; the residual ambiguity (a ≥2-sample blip inside a <400 ms stroke splits it) is a deliberate tradeoff, documented in the code.
4. **Sticky cooldown suppression** — a press starting during the post-reaction cooldown and held past its expiry used to emit anyway.
5. **Constants**: press confirm 2→1 sample, cooldown 800→250 ms, reaction hold 3000→1200 ms.

AGENTS.md tables and both READMEs' known-issues entries updated to match.

Status of verification: the sensitivity values were tuned live on production hardware; the state-machine changes are compile-verified (esp32s3, `release.py stackchan`) and walkthrough-verified with sample-by-sample timelines. I'm flashing them onto my unit today and will report the on-device results (incl. a rapid five-tap sequence) on this PR.


tsuru0805:
On-device results (production unit, firmware built from this branch):

- **Idle false-positive gate**: 20 samples over 10 s with nobody touching — **0 hits** (CH1 at 0.90 %, other channels at 0.50 %).
- **Rapid-tap sequence**: 6 taps in ~3 s, **all six registered individually and all classified TAP** with realistic measured durations (100–299 ms). On the previous firmware the release debounce inflated every gesture past 400 ms (so quick taps classified as STROKE), and the second tap of a rapid sequence was merged into the first press.
- Boot log shows the new sensitivity read-back working: `Si12T: init OK: ctrl=0x03 out1=0x00 sens12=0x02 sens34=0x00`.
- Firmware serial log and gateway event log agree on all six events (`touch event: TAP ... duration=100 ms` ↔ `touch/tap action=head_pat duration=100ms`).

Happy to adjust the per-channel defaults if other units' coupling differs — the read-back line makes that diagnosable from the boot log.

kisaragi-mochi (changes requested):
Thank you for the careful hardware work here. I checked the exact head (`cd532114`) against the current main, built it successfully with the canonical `release.py stackchan` path, and reviewed the touch state machine with sample-by-sample timelines. The sensitivity register packing itself looks consistent with the Si12T/TSM12 layout, but I found two user-visible paths that need another pass before merge.

### 1. The new cooldown creates a deterministic dead zone for ordinary repeated taps

The rapid-split branch accepts a re-press while release debounce is still pending, but the normal falling-edge path starts the 250 ms cooldown only after the fourth no-touch sample confirms release.

At the 10 Hz poll rate, using the first no-touch sample as `t=0`:

- `t=0,100,200`: release is pending.
- `t=300`: the fourth no-touch sample confirms release; cooldown is then set until `t=550`.
- A re-press around `t=200..300` takes the rapid-split path and is accepted.
- A re-press around `t=400..500` occurs after release confirmation but during that delayed cooldown, so `touch_press_suppressed_` drops it.
- A later re-press is accepted again.

That makes the behavior non-monotonic: a roughly 300 ms gap works, a roughly 400-650 ms gap can be swallowed depending on sampling phase, and a slower tap works again. The reported rapid sequence exercises the split path, so it does not cover this middle-gap path.

Please make the normal and split paths use the same physical-release basis for cooldown (for example, derive it from `touch_pending_since_us_`, not the debounce-confirmed `now_us`) and add timeline coverage around 300/400/500/600 ms gaps so accepted tap counts stay monotonic.

### 2. Fresh-reset CH2/CH3 are not enabled/calibrated

`Si12T::Begin()` now writes sensitivity registers `0x02/0x03` and `CTRL2`, but it still does not initialize the channel/reference/calibration registers at `0x0A..0x0F`. The Si12T reset state leaves channels beyond CH1 held/disabled; M5Stack's official StackChan Si12T setup explicitly writes those six registers to enable sensing and calibration before reading the three-zone result.

This driver advertises and ORs CH1-CH3, and both READMEs now describe three zones. On a fresh power-on, however, touches coupled primarily to CH2/CH3 can remain invisible. The PR's six-tap log demonstrates tap classification, but it does not include `start_zones`/`start_raw` evidence for all three zones, so it can still be a CH1-only test.

Please initialize the three used channels (including reference/calibration state) and verify each physical zone separately with the raw channel log.

### Follow-ups while updating

- Add the required `[Unreleased] > ### Firmware` CHANGELOG entry; this repository requires one for firmware source changes.
- `firmware/main/boards/stackchan/AGENTS.md` still says touch reactions last 3 seconds in the layer table, while the new constant is 1200 ms.
- The sensitivity readbacks currently ignore `SafeReadReg()` failures and can print `init OK` with zero-filled diagnostic values. Please either fail initialization or make the log explicitly show readback failure.

Validation on my side: `git diff --check` passed; the canonical ESP-IDF 5.5.2 StackChan build passed and produced a 0x2e6310-byte app image with 26% of the app partition free. I did not flash or claim maintainer-side hardware validation.

Issue #298, opened 2026-06-18
Audio sensing — surface touch-initiated listen transcripts as Channels notifications


kisaragi-mochi:
Parent: #297

## Background

The firmware ships a touch-driven listen UX (PR #208) where a tap on
the head starts a single-shot push-to-talk listening session. The
companion notification path for the touch tap itself — Channels
notifications via the `stackchan/event` capability — is already in
place (#260, closed). What is missing is the audio side: when the
touch-initiated listening session produces a transcript, it should
surface in the controlling agent's context as a `stackchan/event`
Channels notification, the same way the touch tap itself does.

Today the upstream agent only sees the transcript if it explicitly
called `listen()` and waits for the response. A device-initiated
listen (= touch triggered) leaves the transcript unrouted from the
agent's perspective.

## Goal

Add a `stackchan/event` notification path so that when a
touch-initiated listen session completes, the recognized transcript
surfaces in the agent's context as a Channels notification — without
the agent having to call `listen()` first.

## Architecture sketch

- Reuse the existing Channels notification infrastructure (the
  `stackchan/event` capability landed in #260) — add a new event
  subtype for `audio.transcript`
- Trigger source: completion of a touch-initiated listen session
  (the same code path that PR #208 introduced)
- Payload: transcript text, capture timestamp, capture duration,
  source = touch
- Opt-in at the gateway level so an existing call-site that already
  reads the transcript via `listen()` does not get a duplicated
  delivery

## Coexistence

- Non-touch-initiated `listen()` calls retain their current
  response-based delivery (= no Channels notification on top, to keep
  programmatic flows untouched)
- The transcript notification path stays additive; the existing
  `listen()` MCP tool surface does not change

## Out of scope

- Always-on VAD (= activation gesture is touch, not voice)
- Wake-word-initiated listen (= explicitly excluded by the umbrella
  non-goals)
- STT accuracy improvements (= tracked under sibling issue)
- Audio bytes themselves over the Channel (= text only, mirroring the
  touch-event pattern from #260)

## Open questions for the implementation phase

- Notification rate-limit / suppression rules to avoid context spam
  when the device is in a noisy environment
- Should the `audio.transcript` notification include the
  device-decided capture duration / capture-end reason for diagnostics?
- Relationship with #233 (= 30s auto-stop) — does the auto-stop need
  to emit a distinct event subtype (e.g. `audio.transcript_timeout`)?

PR #376, opened 2026-09-02
feat: factory first-run docs, classic avatar, and connect-time gateway hooks

iammikek:
## Summary
- Factory-kit → Cursor first-run guides (`docs/first-run-from-factory.md` / `.ja.md`), linked from README and `AGENTS.md`.
- Classic Stack-chan avatar example plus connect-time `STACKCHAN_AVATAR_SET_PATH` autoload (exact layered/matrix size only; unknown sizes fail closed).
- When an avatar set is configured, the same hook re-enables blink after hello.
- `STACKCHAN_LISTEN_LANGUAGE` for `listen()` when language is omitted; MCP schema no longer advertises a language default.

## Review follow-up
Addressed CHANGES_REQUESTED on this PR:
- Removed `STACKCHAN_AUDIO_HOOK_URL=local` / in-process parrot and `examples/audio-hook-receiver/` from this PR (HTTP hook dump path from #209 remains unchanged).
- Dropped `listen` schema `language` default so Cursor does not force `ja` over the env.
- Avatar size inference fails closed unless mode is set or size matches layered/matrix exactly.
- Blink-on is gated on avatar autoload (not every connect).
- Changelog Docs now records the first-run guides.

## Test plan
- [x] `cd gateway && uv run --group dev pytest` (focused suites) + ruff on touched files
- [ ] Manual: set `STACKCHAN_LISTEN_LANGUAGE=en`, omit language in Cursor `listen()`, confirm English STT
- [ ] Manual: classic avatar path autoload + blink after RST
- [ ] Manual: unknown-size avatar path without mode → skipped with warning


kisaragi-mochi (changes requested):
Thanks for the first-run writeup — the unbind / phone-Wi-Fi / `--no-mdns` / Advanced URL path is the part that actually helps someone with a factory kit.

Please split the new product feature out of this PR, and treat the rest as requested changes rather than follow-ups.

## Split out: in-process tap-to-talk parrot

`STACKCHAN_AUDIO_HOOK_URL=local` (gateway transcribes a device-driven tap and `say`s `You said: …`) is a new feature, not first-run setup.

What already exists is only the opt-in HTTP dump from #209: tap/wake-word audio is POSTed as Ogg/Opus when a URL is set, otherwise discarded. The gateway does not transcribe or speak. Cursor `listen()` / `say()` already cover a desk smoke test.

Please move out of this PR:

- `STACKCHAN_AUDIO_HOOK_URL=local` and `handle_local_capture` / `You said:`
- `examples/audio-hook-receiver/`
- first-run §7c and the matching changelog / `.env.example` / CLI help for that path

This also is not #298 (touch transcripts as Channels notifications to the agent). A parrot with no LLM does not satisfy that issue.

If you open that as its own PR, please also fix there:

- Fire-and-forget STT+`say` on tap stop: overlapping taps can talk over each other.
- `examples/audio-hook-receiver/` defaults language to `en` while the gateway listen default is `ja`.

## Requested changes on what stays

**1. `STACKCHAN_LISTEN_LANGUAGE` vs the `listen` schema default**

`listen` still advertises `"default": "ja"`. Cursor often sends that default even when the tool call omitted `language`, so `STACKCHAN_LISTEN_LANGUAGE=en` never applies on the MCP path. The first-run English-listen section will not work from Cursor as written.

Drop the schema default (or document that clients must omit `language`), and make the description match the resolve order: explicit argument, then env, then `ja`.

**2. Unknown avatar-set size must not become `matrix`**

`resolve_configured_avatar_set()` treats any non-layered size as matrix. A truncated or wrong layered file then fails as a matrix size mismatch. If `STACKCHAN_AVATAR_SET_MODE` is unset and the size is not exactly layered or matrix, fail closed and log; do not guess.

**3. Blink-on is a behavior change for every connect**

`_enable_connect_blink` runs even when `STACKCHAN_AVATAR_SET_PATH` is unset. That turns blink on for all gateway users after hello, not only classic-face setups. Firmware starting with blink off does not by itself mean the gateway should force it on. Gate it on the avatar autoload path, or make it a separate opt-in env, and changelog it as a behavior change.

**4. Changelog the first-run docs**

`CHANGELOG.md` `[Unreleased]` Docs should mention `docs/first-run-from-factory.md` / `.ja.md` (and the README / AGENTS.md links). Right now the Docs subsection does not record that page.

Happy to re-review once the parrot path is out and the items above are addressed.

iammikek:
Addressed the review:

1. **Split out the parrot** — removed `STACKCHAN_AUDIO_HOOK_URL=local`, `handle_local_capture`, first-run §7c, and `examples/audio-hook-receiver/` from this PR. The existing HTTP Ogg dump path (#209) is unchanged.
2. **`STACKCHAN_LISTEN_LANGUAGE`** — removed the MCP `listen` schema `language` default and documented resolve order (arg → env → `ja`).
3. **Avatar size** — unknown sizes / missing file without an explicit mode now fail closed with a warning instead of guessing `matrix`.
4. **Blink** — gated on avatar autoload (`STACKCHAN_AVATAR_SET_PATH` set); no longer forced on for every connect. Changelog calls out the behavior change.
5. **Changelog Docs** — records `docs/first-run-from-factory.md` / `.ja.md` and the README / AGENTS links.

Pushed in `46937bd`. Happy to open the local tap-to-talk path as a separate PR later if useful.

